PR #366, opened 2014-05-24
Infinitescroll evaluating response with javascript

### LINKED ISSUE #363: InfiniteScrollingBehaviour does not evaluate JavaScript in Ajax-Respone
See https://github.com/gesellix/wicket-bootstrap/compare/infinite-scroll-with-javascript-response for an example using InfiniteScrollingBehaviour together with a PopoverBehaviour. It is expected that one can click on any Item in the infinite scrolling list to make a popover visible.

It seems like Wicket doesn't handle the <evaluate> element from subsequent responses. Do I need to manually trigger an evaluation (how?) or wouldn't it be better to make the infinite scrolling component integrate more tightly with Wicket?

I already tried to modify `jquery.infinitescroll.js` in the `_loadcallback` function, but I don't know much about the mechanics and how to integrate Ajax correctly with Wicket. The Wicket docs didn't help very much. If you could give me some hints (and fixing the InfiniteScrollingBehaviour would take too long), I could continue trying to fix this issue.

Thanks!

gesellix:
Infinitescroll wasn't aware of Wicket specific ajax-response processing. This PR adds a slightly modified Wicket.Ajax.process(data) call to the append handler of subsequent page requests.

fixes #363

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/navigation/InfiniteScrollingBehavior.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/navigation/InfiniteScrollingBehavior.java
@@ -22,6 +22,7 @@ import java.util.Map;
  * @link http://www.infinite-scroll.com/
  */
 public class InfiniteScrollingBehavior extends Behavior {
+    private static final ResourceReference WICKET_ADAPTER = new JQueryPluginResourceReference(InfiniteScrollingBehavior.class, "js/infinitescroll-wicket-adapter.js");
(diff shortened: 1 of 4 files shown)


martin-g:
@gesellix Do you still need this changes?

gesellix:
@martin-g oops, no, thanks for checking!
Out of curiosity: is there a better solution now?

martin-g:
Honestly, I don't even remember what this JS plugin does exactly.

In my current project I use DataTables.net JS widget to provide "virtual scroll" functionality to an HTML `table` element. But I have the feeling this is something different.

gesellix:
I'm also unsure about the real difference, but if I recall correctly this plugin is a bit more generic. Since I'm actually in the process of migrating my site away from Wicket, this won't be necessary anymore. Thanks again for the little cleanup of such an old issue.

PR #483, opened 2014-12-29
Better styling for BootstrapPagingNavigator, using Bootstrap `class="disabled"`

ceefour:
Better styling for BootstrapPagingNavigator, using Bootstrap `class="disabled"`

PR #485, opened 2015-01-02
Better styling for BootstrapPagingNavigator, using Bootstrap `class="disabled"`

ceefour:
Better styling for BootstrapPagingNavigator, using Bootstrap `class="disabled"`

Replaces #483.


ceefour:
Travis build still fails, but it doesn't seem to be related with this pull request.

Anyway, next is rel=prev/next (https://issues.apache.org/jira/browse/WICKET-5799)

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/navigation/BootstrapPagingNavigator.java
```diff
@@ -60,6 +63,39 @@ public BootstrapPagingNavigator(final String markupId, final IPageable pageable,
...
+                        return get("next").isEnabled() ? null : "disabled";
+                    }
+                })));
+        add(new TransparentWebMarkupContainer("lastItem")
+                .add(new CssClassNameAppender(new AbstractReadOnlyModel<String>() {
+                    @Override
+                    public String getObject() {
+                        return get("last").isEnabled() ? null : "disabled";
+                    }
+                })));
```
martin-g: Create a new component (private static class) for this need instead of copy/pasting the same code 4 times.

PR #489, opened 2015-01-23
Provide access to LessCompiler configuration

emlo:
Make it possible to provide a custom LessCompiler.Configuration with application specific data. For example to register custom less functions which can be used to change CSS properties at runtime.


martin-g:
Thanks!
I'll check it soon!

martin-g:
Very good quality of the PR!
Thank you!

PR #499, opened 2015-03-20
animate.css

danjee:
Added animate.css to wicket-bootstrap

### DIFF
--- a/bootstrap-core/pom.xml
+++ b/bootstrap-core/pom.xml
@@ -40,6 +40,12 @@
             <artifactId>modernizr</artifactId>
             <version>${modernizr.version}</version>
         </dependency>
+        
+        <dependency>
+		    <groupId>org.webjars</groupId>
+		    <artifactId>animate.css</artifactId>
+		    <version>${animate.css.version}</version>
+		</dependency>
 
         <dependency>
             <groupId>joda-time</groupId>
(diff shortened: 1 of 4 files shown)


martin-g:
Thank you!

martin-g:
We need to add a demo of it in the -samples project.

tahitiangabriel:
Shouldn't it be in bootstrap-extensions rather than in bootsrap-core?

martin-g:
Ah, I didn't notice this!
Right. It should be in -extensions.

danjee:
I've added in the core next to the BadgeBehavior, QuoteBehavior and other Behavior implementations. 
I will also add a demo for it

martin-g:
-core contains only things coming from http://getbootstrap.com/.
-extensions is for third party addons.
Animate.css is not Bootstrap related but I think it is OK to have here because it could make the components even more awesome.

danjee:
I see. I will make only modifications in the samples pages then.

tahitiangabriel:
If it stays in the core you are adding a dependancy to org.webjars.animate.css which most people won't use.
And core dev would have to mainain it for ever! ;-)
As it is a third party addon it should be in the extension module. It would be cleaner.
My 2 cents.

martin-g:
@danjee Please move it from -core to -extensions if you have time.
Otherwise I'll do it the next days.

danjee:
Ok, I will move it in the next commit with samples.
Thanks Martin and Gabriel

PR #500, opened 2015-03-20
Move component to extensions and add sample for it

danjee:
move behavior from core to extensions including test class
added sample with notification panel animation


martin-g:
Thanks!

danjee:
Yoa are welcome!

PR #505, opened 2015-04-09
Typeahead.js fixes

pulse00:
- Having a local AND a remote option is valid
- Added the StalePageException fix from @osterberg
- Fixed template handling


martin-g:
Thanks!

Issue #371, opened 2014-05-29
DropZone - modern file uploader


martin-g:
It seems http://www.dropzonejs.com/bootstrap.html provides what http://blueimp.github.io/jQuery-File-Upload/ was missing (see http://wicketinaction.com/2012/11/uploading-files-to-wicket-iresource/).

Do we want integration with it ?
Please vote if you want it integrated.

P.S. The minimal IE version that it supports is IE10.

damienhollis:
+1 this would be great

klopfdreh:
I am working on it and create a pull request when it's done.

klopfdreh:
See also the following commits:

https://github.com/klopfdreh/wicket-bootstrap/commits/master

and a request to allow multipart in AbstractDefaultAjaxEventBehavior:

http://apache-wicket.1842946.n4.nabble.com/Change-AbstractDefaultAjaxBehavior-to-allow-multipart-td4670748.html

klopfdreh:
Hi Martin - I think this can be closed, now. :-)

martin-g:
Thank you for the contribution, Tobias!

PR #508, opened 2015-05-14
Modern File Upload based on dropzone.js

klopfdreh:
https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/371

### DIFF
--- a/bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/pages/ExtensionsPage.java
+++ b/bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/pages/ExtensionsPage.java
@@ -21,10 +21,7 @@ import org.apache.wicket.markup.html.link.AbstractLink;
 import org.apache.wicket.markup.repeater.RepeatingView;
 import org.apache.wicket.model.Model;
 import org.apache.wicket.model.ResourceModel;
-import org.apache.wicket.protocol.http.servlet.MultipartServletWebRequest;
-import org.apache.wicket.protocol.http.servlet.ServletWebRequest;
 import org.apache.wicket.request.mapper.parameter.PageParameters;
-import org.apache.wicket.util.lang.Bytes;
 import org.apache.wicket.util.time.Duration;
 import org.apache.wicket.util.upload.FileItem;
 import org.wicketstuff.annotation.mount.MountPath;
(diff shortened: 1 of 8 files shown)


klopfdreh:
bildschirmfoto 2015-05-14 um 11 56 07

bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/pages/ExtensionsPage.java
```diff
@@ -204,20 +194,47 @@ protected WebMarkupContainer createDialog(String id) {
...
+			@Override
+			protected void onUpload(AjaxRequestTarget target)
+			{
+				try
+				{
+					ServletWebRequest webRequest = (ServletWebRequest)getRequest();
+					MultipartServletWebRequest multiPartRequest = webRequest.newMultipartWebRequest(
+						Bytes.bytes(getMaxFileSize() * 1000000), "ignored");
+					multiPartRequest.parseFileParts();
+					Map<String, List<FileItem>> files = multiPartRequest.getFiles();
```
martin-g: I think the parsing code should be in the Ajax behavior. 
The signature of `#onUpload()` should be `#onUpload(AjaxRequestTarget, Map<String, List<FileItem>>)`

martin-g:
Thank you, Tobias!
I'll merge the PR soon. There are some small issues which I want to address (code style, JSON config) right after the merge.

bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/pages/ExtensionsPage.java
klopfdreh: Okay - we can do this - I also thought about it but I decided to handle it like in wicket forms. I am fine with this, too.

klopfdreh:
Ah maybe I create a pull request for wicket-eclipse-settings - then the code style is no topic to discuss anymore. ;-)

Thank you for the fast feedback!

martin-g:
I think @l0rdn1kk0n would prefer to keep the current format settings.

klopfdreh:
Okay - no problem - then I am going to setup my IDE to fit those settings if I contribute to wicket-bootstrap.

klopfdreh:
I parsed the file parts internally, now - I wrapped the file upload exception into a wicket runtime exception - hope this is ok.

klopfdreh:
Ah maybe we can add a method onError which is called in the catch block of response and takes the Exception and the AjaxRequestTarget as argument.

martin-g:
I've reworked a bit your work in branch '508-dropzone'.
And now it is broken :-)
I'll continue tomorrow.

klopfdreh:
It is not working but the the config thing is really cool! I think of rework some html5 commits in wicketstuff with that.

klopfdreh:
I solved the issue why it is not working - I create a pull request. :-)

damienhollis:
@martin-g I looked at what @klopfdreh was working on for this last week and noticed that there was no wicket-ajax support.  I was hoping you guys might have a creative way to make it work but obviously it isn't possible.  However, I was wondering whether it would be possible to hook into dropzone's success event and cause a subsequent Ajax request but this time using wicket.  This would allow us to update the state of the page after a file was successfully uploaded.  

Is this possible?  Is it worth doing or do you have a better proposal?

martin-g:
@damienhollis Yes. This should be possible. Please open a new ticket.

Wicket-Atmosphere works in a similar way: https://github.com/apache/wicket/blob/6dfe99f17a989c9d9759cd6ce2830f7962a9090d/wicket-experimental/wicket-atmosphere/src/main/java/org/apache/wicket/atmosphere/jquery.wicketatmosphere.js#L30

klopfdreh:
http://www.dropzonejs.com/#events

complete all files haven been uploaded / success one file has been uploaded.

Do you want to react on each file or on complete? 

On each file the AjaxRequestTarget can be used - or am I missing something?

martin-g:
This is what I read at there docs:

event name | description
-----------------|---------------
success | The file has been uploaded successfully. Gets the server response as second argument. (This event was called finished previously)
complete | Called when the upload was either successful or erroneous.

It doesn't say that `complete` is after all. 
It looks like jQuery's behavior: first calls either `success` or `canceled` and then calls `complete`.

IMO you should listen for `complete` and if it contains `<ajax-response ` then try to process it.

klopfdreh:
I had a look into the response of the upload post and there is only a complete html content shipped as response but no &lt;ajax-response&gt; - maybe we have to do something on server side with the response first to be able to process it?

martin-g:
Si, señor! :-)
We have to add `AjaxRequestTarget` back in the signatures of `#onUpload` and `#onError` and write something back to the client.

klopfdreh:
Aha! Gotcha! ;-)

klopfdreh:
Client side is clear (maybe add some check if server side send 200):

myDropzone.on("success", function(x,response) {
	Wicket.Ajax.process(response.responseBody);
});

But for the response on server side I don't know - there something like a wicket response builder to be used within the DropZoneFileUploadAjaxEventBehavior?! :-)

martin-g:
What do you mean?
The user should just write to the AjaxRequestTarget

klopfdreh:
That was not working for me - try it out - a whole html response was providet but no ajax-response. :-:

damienhollis:
@klopfdreh I haven't looked at your latest code, so you might already be doing what I'm about to suggest.  What I noticed while trying to get dropzone working with wicket was that wicket was treating dropzones ajax posts as if they are regular posts and not ajax.  What I had to do was add a header to the post that told wicket to treat the post as ajax.  The header I added was "Wicket-Ajax" - have a look at what WebRequest.isAjax() does.  I added the header by using this option in dropzone: http://www.dropzonejs.com/#config-headers

Hopefully that helps.

klopfdreh:
var config = ${config};
config.headers = {"Wicket-Ajax":"true","Wicket-Ajax-BaseURL":Wicket.Ajax.baseUrl}
var myDropzone = new Dropzone(document.body,config);

I receive the response, now. I try to finish it.

klopfdreh:
Ok, got it - I replaced a label with the response. I am going to create a pull request.

klopfdreh:
There are some things to know: 

1. I store the response of one of the file posts and only process it once in queuecomplete, otherwise the content would be replaced for each file
2. One ugly thing left is that the response is shipped for each file so it is not so good to replace large area of content

@damienhollis thanks a lot for the hints - I was already digging around that area with the wicket ajax request, but I didn't know the header name :-)

PR #509, opened 2015-05-16
Fixed Dropzone with new Config (508-dropzone)

klopfdreh:


klopfdreh:
Meeee I destroyed the formatting :( - Can you give me a hint which formatter is used in wicket-bootstrap?

klopfdreh:
Thank you! Good to know about Json.RawValue.

I saw that the TemplateMechanism is wicket-bootstrap related, but I think it can be used in MANY situations. Would you mind to integrate it in wicket-core?

martin-g:
We try to keep Wicket having less dependencies.
wicket-jquery-selectors project depends on Jackson for JSON generation.

Extra reason not to add dependency to Jackson: https://github.com/l0rdn1kk0n/wicket-jquery-selectors/issues/11
RawValue is broken between Jackson 2.3.0 and 2.6.0 (not released yet).

klopfdreh:
Oh I saw that it is in wicket-core but not IKey and Json - so let me adress my question right: Would you mind to integrate IKey and Json into wicket-core?!

PR #511, opened 2015-05-20
Added a generic method for option in DatetimePickerConfig

VigneshPalanisamy:
Added a generic method for option in DatetimePickerConfig
Updated the Eonasdan-bootstrap-datetimepicker version to 4.7.14

### DIFF
--- a/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePickerConfig.java
+++ b/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePickerConfig.java
@@ -245,6 +245,18 @@ public class DatetimePickerConfig extends AbstractConfig {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePickerConfig.java
```diff
@@ -245,6 +245,18 @@ public DatetimePickerConfig withEnabledDates(Date[] enabledDates) {
     }
 
     /**
+     * Set Option
+     *
+     * @param key jquery key
+     * @param value new value for the key
+     * @return config instance
+     */
+    public <T> DatetimePickerConfig withOption(IKey<T> key, T value) {
```
martin-g: This is not the right place/way.
This is basically making `#put()` public instead of protected. 
We intentionally made it protected so that you list the available options as setter methods with fluent API.
With this method it will be possible to set non-sense options.

PR #513, opened 2015-05-31
JDK8 junit fix - wicket-jquery-selector 0.1.5 has to be released first!

klopfdreh:


klopfdreh:
Yep build failed because wicket-query-selector 0.1.5 isn't released.

klopfdreh:
prerequisite: https://github.com/l0rdn1kk0n/wicket-jquery-selectors/pull/14

PR #534, opened 2015-07-21
Add missing FontAwesome database icon

mrijke:
The FontAwesome [database icon](http://fortawesome.github.io/Font-Awesome/icon/database/) was missing from the Java classes but present in the css. This commit adds it.


l0rdn1kk0n:
thanks.

Issue #512, opened 2015-05-20
Add integration with a decent WYSIWYG editor


martin-g:
E.g. https://github.com/summernote/summernote

klopfdreh:
Hi,

just to let you know that I tried to integrate it, but there were a lot of styling issues with the current theme, which makes the editor not usable at the current state. Maybe I did something wrong, but as of the documentation it is rather easy to integrate it with:

```java
$(document).ready(function() {
  $('#summernote').summernote();
});
```

martin-g:
So Summernote looks well only with the default Bootstrap theme? And looks bad with any other theme (e.g. Bootswatch, Material design)?

klopfdreh:
I only tested it with the theme of the wicket-bootstrap-examples (in which I integrated the DropZoneFileUpload) and in this theme the layout was broken. Maybe it was only related to this theme, but I just wanted to give a short feedback. I didn't test int with Bootswatch / Material

martin-g:
The samples run with Bootswatch theme by default. There is a theme chooser at the top-right corner.
To change the theme provider (to Bootstrap, Material Design or Vegibit) you have to edit WicketApplication.java directly.

I'll take a look soon.
Thanks, Tobias!

klopfdreh:
I managed to get it working - I am going to finalize it soon.

klopfdreh:
<img width="960" alt="summernote" src="https://cloud.githubusercontent.com/assets/980773/8899273/906b406c-3432-11e5-8a2d-0a1f1a0fcbed.png">

klopfdreh:
One point we should discuss is how to get the content of the editor shipped to the backend. Maybe via button or a hook to embed the editor into a form? What do you think?

reiern70:
Hi,

I haven't looked at your integration but... Maybe use the call-backs 
http://summernote.org/#/deep-dive#callbacks

to get contents into a hidden input? E.g

1- Roll out a behaviour creating client side-component. 
2- An "TextField" component that creates the a hidden input + client side so that text is synchronized with hidden field when it changes.
3- You will need to add server side call-back to upload images and so on?

martin-g:
Hi,

I haven't looked neither at Tobias' work nor at Summernote docs but I'd expect the editor to submit its text as a normal FormComponent, i.e. if it is a part of a Form then submitting the form should update the editor's model object. Also AjaxFormComponentUpdatingBehavior should work as for any other FormComponent.

[...]

PR #536, opened 2015-08-09
Add integration with a decent WYSIWYG editor

klopfdreh:
https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/512


bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.html
```diff
@@ -0,0 +1,10 @@
+<html xmlns:wicket="http://wicket.apache.org">
+	<head>
+		<title>SummernoteEditor</title>
+	</head>
+	<body>
+		<wicket:panel>
+			<div id="summernote"></div>
```
martin-g: This hardcoded id looks problematic.
You cannot have more than one editor per page this way.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
```diff
@@ -0,0 +1,160 @@
...
+
+	@Override
+	protected void updateAjaxAttributes(AjaxRequestAttributes attributes) {
+	    super.updateAjaxAttributes(attributes);
+	    attributes.setMultipart(true);
+	}
+    }
+
+    public SummernoteEditor(String id) {
+	this(id, null, new SummernoteConfig());
```
martin-g: Broken formatting.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.wicket.WicketRuntimeException;
+import org.apache.wicket.util.lang.Args;
+
+import de.agilecoders.wicket.jquery.AbstractConfig;
+import de.agilecoders.wicket.jquery.IKey;
+
+public class SummernoteConfig extends AbstractConfig {
```
martin-g: Please add a javadoc with a link to Summernote's options documentation.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private String subFolderName;
+
+    /**
+     * @param airmode
+     *            if air mode should be used or normal
+     * @return current instance
+     */
+    public SummernoteConfig withAirMode(boolean airMode) {
+	put(AirMode, airMode);
```
martin-g: Mixed tab/spaces

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
```diff
@@ -0,0 +1,166 @@
...
+	return get(MaxFileSize);
+    }
+
+    /**
+     * @param callbackUrl
+     *            The image upload url to call back
+     * @return current instance
+     */
+    public SummernoteConfig withImageUploadCallbackUrl(String callbackUrl) {
+	put(ImageUploadCallbackUrl, callbackUrl);
```
martin-g: This is asking for troubles.
Usually the callback urls should not be stored.
They have to be generated at RENDER phase and immediately written to the HTML response.
If stored for later use than there is a (big) chance that they will become obsolete, e.g. the page's `renderCount` or behavior id will change.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
```diff
@@ -0,0 +1,166 @@
...
+     * 
+     * @param uploadFolder
+     *            the upload folder
+     * @param subFolderName
+     *            the name of the sub folder within the upload folder
+     */
+    public static void setUploadFolder(File uploadFolder, String subFolderName) {
+	uploadFolder = Args.notNull(uploadFolder, "uploadFolder");
+	if (!uploadFolder.isDirectory()) {
+	    throw new WicketRuntimeException("The given file is not a folder" + uploadFolder);
```
martin-g: Add a space after 'folder'.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
```diff
@@ -0,0 +1,166 @@
...
+	}
+	File subFolder = new File(uploadFolder, subFolderName == null ? "" : subFolderName);
+	if (!subFolder.exists()) {
+	    boolean created = subFolder.mkdirs();
+	    if (!created) {
+		throw new WicketRuntimeException("folder structure could'nt be created" + subFolder.getPath());
```
martin-g: couldn't

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * 
+ * Please add
+ * 
+ * mountResource(SummernoteStoredImageResourceReference.SUMMERNOTE_MOUNT_PATH,
+ * new SummernoteStoredImageResourceReference(new File("/Users/MyUser/"),"subpath"));
+ * 
+ * and then use config
+ * SummernoteConfig config = new SummernoteConfig();
+ * config.setSubFolderName(subpath");
```
martin-g: missing opening "

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
```diff
@@ -0,0 +1,160 @@
...
+	super(id, model);
+	this.config = Args.notNull(config, "config");
+	add(summernoteEditorImageAjaxEventBehavior = new SummernoteEditorImageAjaxEventBehavior());
+    }
+
+    @Override
+    public void renderHead(IHeaderResponse response) {
+	response.render(JavaScriptHeaderItem.forReference(SummernoteEditorJavaScriptReference.instance()));
+	response.render(CssHeaderItem.forReference(SummernoteEditorCssReference.instance()));
+	response.render(CssHeaderItem.forUrl("//netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.min.css"));
```
martin-g: Please use FontAwesomeCssResourceReference instead.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditorCssReference.java
```diff
@@ -0,0 +1,27 @@
+package de.agilecoders.wicket.extensions.markup.html.bootstrap.editor;
+
+import org.apache.wicket.request.resource.CssResourceReference;
+
+public class SummernoteEditorCssReference extends CssResourceReference {
```
martin-g: Can we use Webjars?
http://search.maven.org/#artifactdetails%7Corg.webjars.bower%7Csummernote%7C0.6.16%7Cjar
I am not sure because of the `.bower` part in the groupId.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditorJavaScriptReference.java
```diff
@@ -0,0 +1,39 @@
...
+     * Private constructor.
+     */
+    private SummernoteEditorJavaScriptReference() {
+	super(SummernoteEditorJavaScriptReference.class, "js/summernote.js");
+    }
+
+    @Override
+    public Iterable<? extends HeaderItem> getDependencies() {
+	ArrayList<HeaderItem> dependencies = new ArrayList<HeaderItem>();
+	dependencies.add(JavaScriptHeaderItem.forReference(JQueryResourceReference.get()));
```
martin-g: It seems you need to extend from JQueryPluginResourceReference and have this dependency for free.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/js/summernote_init.js
```diff
@@ -0,0 +1,34 @@
+$(function() {
+	var summernoteconfig = ${summernoteconfig};
+	var summernoteconfigdefault = {
+		onImageUpload : function(files) {
+			$(files).each(function(){
+				var file = this;
+				var data = new FormData();
```
martin-g: According to http://caniuse.com/#search=FormData only IE Edge supports FormData without issues.
We may need to use iframes for the uploads instead.

martin-g:
I've added some comments on the code itself.
I'll have to play with the demo (where is it?!) and take a deeper look at the logic for file/image uploads to be able to say whether I see any improvements.

klopfdreh:
There is no demo at the moment, because I thought it is to early to put it in. But to make it short:

WicketApplication (last line in init):
```java
mountResource(SummernoteStoredImageResourceReference.SUMMERNOTE_MOUNT_PATH, new SummernoteStoredImageResourceReference(new File("/Users/klopfdreh/"),"test"));
```

ExtensionPage:
```Java
    Form<Void> form1 = new Form<Void>("form1");
    
    SummernoteConfig summernoteConfig = new SummernoteConfig();
    summernoteConfig.setSubFolderName("test");
    form1.add(new SummernoteEditor("summernote",Model.of("sxa"),summernoteConfig));
    form1.add(new AjaxButton("submit") {
	private static final long serialVersionUID = 1L;
    
        @Override
        protected void onSubmit(AjaxRequestTarget target, Form<?> form) {
    		System.out.println("sdsc");
        }
    }.setDefaultFormProcessing(false));
    add(form1);
```

ExtensionPage.html
```html
<section id="su">
	<form wicket:id="form1">
		<div wicket:id="summernote"></div>
		<input wicket:id="submit" type="submit" value="submit"/>
	</form>
</section>
```

klopfdreh:
The submit to send the content of the editor to the server is not implemented yet.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
klopfdreh: Then we should also change this for the DropZoneFileUpload:

https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/master/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/fileUpload/DropZoneConfig.java#L37

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
klopfdreh: The documentation is not finished in general - if the code is ok we should finish it. I only made some short hints, because a lot of changes were done and the doc changes a lot, too

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/js/summernote_init.js
klopfdreh: is there any other way to transfer data without using iframe - I dislike that hack. :tired_face:

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.html
klopfdreh: Good point - how should we make it configurable?

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
klopfdreh: Will be done if the API is finished.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
klopfdreh: I hate my formatter - this happens by time...

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
klopfdreh: Thanks!

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditorJavaScriptReference.java
klopfdreh: Done.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
klopfdreh: Done.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditorCssReference.java
klopfdreh: DropZoneFileUpload also stores the library locally and I think as long as the implementation doesn't change we can load it from the applications server.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
klopfdreh: Done.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
```diff
@@ -0,0 +1,166 @@
...
+     * Sets the name of the sub folder
+     * 
+     * @param subFolderName
+     *            the name of the sub folder
+     */
+    public void setSubFolderName(String subFolderName) {
+	subFolderName = Args.notNull(subFolderName, "uploadFolder");
+	this.subFolderName = subFolderName;
+    }
+}
```
reiern70: Hi,

Thanks for your work!

Question: Why not make uploading/downloanding interface based? E.g.

IUploadHandler {
   void storeImage(....)
   
   ImageFile getStoredImage(....) 
}

That way the FileSystem is just an option?

klopfdreh: Good point, we should use an interface instead of File - @martin-g I don't know but is there such an interface used in wicket-core already? Maybe we can use that instead of defining our own in that case.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditorCssReference.java
martin-g: The idea of Webjars is that the application can provide custom version of the JS widget (newer or older one) by adding Maven dependency to its pom.xml (same for Gradle, SBT, ...).
By putting the resource in the classpath we should either add custom code to be able to set custom ResRef or one has to deal with ClassLoader ordering to provide custom version of the .js/.css resource.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.html
martin-g: Make it a Wicket component and use its `getMarkupId()` where needed.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/js/summernote_init.js
martin-g: I am not aware of such.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
martin-g: Yes, most probably.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteConfig.java
martin-g: byte[] ? :-)

martin-g: Well, it should be byte[] + MIME type.
I am not aware of such tuple in wicket-core.

klopfdreh:
I'm going to do the following integrations this weekend:
* Summernote via webjars
* User getMarkupId instead of a hard-coded id
* Interface for Streams

klopfdreh:
@martin-g would you be so kind and help me out with the submit of the editor content? I think it has to 
be a JavaScript related implementation, because of the summernote editor.

If the editor is rendered on request we have to do something like:
```Java
// Ajax Request against behavior to get the content
$("#"+markupId).code(content);
```

FormSubmit where I don't know how to hook in into a Wicket Form Submit
```Java
// User has finished editing the content of the editor
var content = $("#"+markupId).code(); // without parameter
// On Form Submit, send the content to the server
```

reiern70:
@klopfdreh 

I do not know if this is the best solution... but is what I remember I did for other editor.

1- Add a hidden field to be submited via wicket AJAX
2- Register an onChange listener that copies $("#"+markupId).code(content); into that hidden field.

klopfdreh:
Okidoki :smiley:

klopfdreh:
The only thing left now is to store the content of the editor and restore it after a page reload. I switched the implementation of File to the interface SummernoteStorage. An example implementation and how to use it can be found here:

https://github.com/klopfdreh/wicket-bootstrap/blob/fd7eaea0dd13c8159add1aa15cfcc18ce9e7870e/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteFileStorage.java

klopfdreh:
To send / receive content is also finished - I used the blur event of the editor which saves the content into a hidden field which is send by form submit. In the backend you can read the content this way:

```java
@Override
protected void onSubmit(AjaxRequestTarget target, Form<?> form) {
    System.out.println(summernoteEditor.getSubmittedContent());
}
```
To apply content to the editor you can do this with the SummernoteConfig

```java
SummernoteConfig.addStorage(new SummernoteFileStorage("fileStorage", new File("/Users/klopfdreh")));
SummernoteConfig summernoteConfig = new SummernoteConfig();
summernoteConfig.useStorageId("fileStorage");
summernoteConfig.withContent("This is the content to be placed initial into the editor");
```

klopfdreh:
The last thing left is:
* Which configurations of the Summernote-Editor should be also available in the Wicket-Bootstrap integration
* Maybe we can parse a raw json object into the config, so that the users can apply very custom settings

martin-g:
> SummernoteConfig.addStorage(new SummernoteFileStorage("fileStorage", new File("/Users/klopfdreh")));
> summernoteConfig.useStorageId("fileStorage");

Why this has to be static + String keys?
Wouldn't be better to be like: `summernoteConfig.setStorage(new SummernoteFileStorage("fileStorage", new File("/Users/klopfdreh"))));`

martin-g:
> summernoteConfig.withContent("This is the content to be placed initial into the editor");

Why not use Summernote's model object as content ?

martin-g:
> Which configurations of the Summernote-Editor should be also available in the Wicket-Bootstrap integration
> Maybe we can parse a raw json object into the config, so that the users can apply very custom settings

I think we can provide a Config class with just a few (the most important according to you) settings. Each application can just extend this config and add any other settings it needs. Usually users just add these extra needed settings to the Config class and send us Pull Request, so they do not maintain custom classes.

klopfdreh:
> Why this has to be static + String keys?
> Wouldn't be better to be like: summernoteConfig.setStorage(new SummernoteFileStorage("fileStorage", new File("/Users/klopfdreh"))));

That is because of the mounted ResourceReference - the id is used of the ResourceReference to get the right storage - it uses a static getter to get the storage. You can initialize a Summernote related reference only with the id and lateron the Storage - see javadoc of SummernoteFileStorage

klopfdreh:
> Why not use Summernote's model object as content ?

Good point I will change it!! :+1:

klopfdreh:
> Each application can just extend this config and add any other settings it needs...

Okay then we only have to decide which configs are most important.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/js/summernote_init.js
klopfdreh: done.

klopfdreh:
Yep, seems to as of http://stackoverflow.com/questions/17626871/formdata-ie-8-does-not-work

klopfdreh:
Ok, I think all required implementations and the JavaDoc are finished, now. Here is an Example:

WicketApplication:
```java
mountResource(SummernoteStoredImageResourceReference.SUMMERNOTE_MOUNT_PATH, new SummernoteStoredImageResourceReference("fileStorage"));
```

Page:
```java
SummernoteConfig.addStorage(new SummernoteFileStorage("fileStorage", new File("/Users/klopfdreh")));

Form<Void> form1 = new Form<Void>("form1");
SummernoteConfig summernoteConfig = new SummernoteConfig();
summernoteConfig.useStorageId("fileStorage");
final SummernoteEditor summernoteEditor = new SummernoteEditor("summernote",Model.of("Test"),summernoteConfig);
form1.add(summernoteEditor);
form1.add(new AjaxButton("submit") {
    private static final long serialVersionUID = 1L;
    
    @Override
    protected void onSubmit(AjaxRequestTarget target, Form<?> form) {
	System.out.println(summernoteEditor.getSubmittedContent());
    }
}.setDefaultFormProcessing(false));
add(form1);
```

Page.html:
```html
<form wicket:id="form1">
	<div wicket:id="summernote"></div>
	<input wicket:id="submit" type="submit" value="submit"/>
</form>
```

@martin-g as you can see, that is the reason why I use the static map - you can register a resource reference only with an id in the application and later on initialize it by calling addStorage. The editor itself gets the hint which storage to use by call useStorage(id)

klopfdreh:
If everything is fine (feedback is welcome :smile:) we can add some examples (maybe that one I provided here).

klopfdreh:
I got one idea last night. If the user doesn't provide a storage, the toolbar item to insert an image should not be available:

http://summernote.org/#/deep-dive#api

klopfdreh:
* New functions addButtons / removeButtons / getButtons of SummernoteConfig
* Removes the picture button if no storage is provided

klopfdreh:
Hi Martin!

Thanks for merging :-)

Maybe we can cherry pick the changes to 0.9.x

In addition we can add some examples to the extension page.

What do you think?

martin-g:
1) master branch is 0.9.x
2) just added a sample page
3) merged to wicket-7.x branch
4) will deploy 7.x at http://wb-mgrigorov.rhcloud.com/ very soon!

klopfdreh:
Thank you very much for the fixes, but there are two little things left to do to enable the upload:

1. set the storage id to the config: myconfig.setStorageId("fileStorage")
2. add a mount for the stored image resource ref in the applications as mentioned in the comment 7 days ago.

This is required so that the editor can access the uploaded images.

martin-g:
http://wb-mgrigorov.rhcloud.com/summernote

I'll do the changes later. Thanks!

klopfdreh:
I always get confused that in wicket bootstrap it is the other way around as in wicket core:

Wicket
master > 7.x
wicket-6.x > 6.x

Wicket Bootstrap
master > 6.x
wicket-7.x > 7.x

martin-g:
Done!

klopfdreh:
Perfect!

* picture button appears
* the image is uploaded
* picture tag appears

klopfdreh:
One bug for mobile devices. We have to add a UUID for the picture names on upload because the name of uploads are always "image.jpg"

klopfdreh:
See https://github.com/klopfdreh/wicket-bootstrap/commit/0280755d5bc16a8eb4907981da2a6195d76e57e5

PR #552, opened 2015-09-12
Summernote as TextArea and addition of BootstrapResourceLink

ymlaine:
Hi, 

I'm still learning git and github, so I finally though that it's better to commit on the master (instead of wicket-7.x branch, am I right ?) and I merged 4 commits in one. So this is a new pull request that replace the yesterday one.

I also added a new BootstrapResourceLink (based on BootstrapLink) that I did in the past, is it ok to put these 2 commits in one pull request ? 

regards,
Yves-Marie

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/BootstrapResourceLink.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/BootstrapResourceLink.java
@@ -1,0 +1,135 @@
+package de.agilecoders.wicket.core.markup.html.bootstrap.button;
+
+import org.apache.wicket.IResourceListener;
+import org.apache.wicket.markup.html.link.Link;
+import org.apache.wicket.markup.html.link.ResourceLink;
+import org.apache.wicket.model.IModel;
+import org.apache.wicket.request.cycle.RequestCycle;
+import org.apache.wicket.request.handler.resource.ResourceReferenceRequestHandler;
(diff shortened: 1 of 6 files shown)


martin-g:
> I also added a new BootstrapResourceLink (based on BootstrapLink) that I did in the past, is it ok to put these 2 commits in one pull request ?

No :-)
Please create separate PR for different contributions.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/BootstrapResourceLink.java
```diff
@@ -0,0 +1,135 @@
...
+import de.agilecoders.wicket.core.markup.html.bootstrap.button.Buttons.Type;
+
+/**
+ * 
+ * Same as default {@link ResourceLink} but based on {@link BootstrapLink} which is styled by bootstrap. 
+ * 
+ * @author ymlaine
+ *
+ */
+public class BootstrapResourceLink extends BootstrapLink<Void> implements IResourceListener {
```
martin-g: I think there is no need of such component in the library.
You can just use: `new ResourceLink(...).add(new ButtonBehavior(...))`.
Copying big chunks of code from Wicket to the library begs for problems in the maintainance in the future

martin-g:
Thanks for the contribution!
The code indeed looks much simpler now!

I've removed BootstrapResourceLink.

klopfdreh:
Thanks from me, too! 

Good to see that the editor is going to be used / improved furtheron! :-)

PR #555, opened 2015-09-17
Addition of BootstrapResourceLink, same as default ResourceLink but b…

ymlaine:
…ased on styled BootstrapLink


martin-g:
Better use Wicket's ResourceLink + Wicket-Bootstrap's ButtonBehavior.

ymlaine:
Hi, I don't think that ResourceLink + Wicket-Bootstrap ButtonBehavior is exactly the same as BootstrapResourceLink, otherwise I'll do that from the beginning. I'm ok that it will add the button behavior, but what about the possibility to add an Icon to the link ? Not a big problem, I will keep it in my workspace, but I really think that this class can help.

PR #568, opened 2015-11-17
Fixes TypeAhead Bloodhound not working with Wicket URL encryption

mrijke:
This commit fixes the Autocomplete TypeAhead Bloodhound class
to work in combination with Wicket's URL encryption.
Originally the getRemoteUrl was malformed, there was no '?'
character denoting the start of the HTTP GET parameters.
Without URL encryption, Wicket happily accepted this malformed URL,
but when URL encryption is enabled it cannot properly decrypt
the malformed URL.


martin-g:
Thank you!

Issue #570, opened 2015-11-21
OptimizedMobileViewportMetaTag is optimized for no zoom


dmbeer:
The are two settings for the viewport tag one is for zoom and mobile devices and the other is without zoom and setting the max scale so that people can't zoom in on the device.

martin-g:
This should be reworked to use either builder pattern of fluent API. I'd prefer Fluent API.
E.g.:
```java
MobileViewportTag mvt = new MobileViewportTag(id);
mvt.width(...).initialScale(...).maxScale(...)
```
Maybe `user-scalable` should be auto-detected if initialScale!=maxScale. (I have to double check that this is its purpose though).

dmbeer:
According to the documentation here http://getbootstrap.com/css/#overview-mobile it suggests that you can only provide it as one of two values but from this blog you can adjust the settings http://www.inmotionhosting.com/support/edu/website-design/create-responsive-template/viewpoint-meta-tag-website-responsive-template.

So actually the way it was correctly done for bootstrap. If however it works by customising it then yes we should make it customisable.

dmbeer:
@martin-g see PR #572

PR #572, opened 2015-11-22
Viewport Meta Tag Refactored #570

dmbeer:
Refactored viewport meta tag to be customisable and to take the parameters required to create the correct viewport tags.


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/html/MobileViewportMetaTag.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.wicket.IGenericComponent;
+import org.apache.wicket.markup.ComponentTag;
+import org.apache.wicket.markup.html.WebMarkupContainer;
+import org.apache.wicket.model.IModel;
+import org.apache.wicket.model.Model;
+
+/**
+ * A special {@link MetaTag} that adds a TODO
+ *
+ * @author miha
```
martin-g: You keep pretending to be `miha` :-)

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/html/MobileViewportMetaTag.java
```diff
@@ -0,0 +1,126 @@
...
+        return name.getObject();
+    }
+
+    public String getWidth() {
+        return width;
+    }
+
+    public void setWidth(String width) {
+        this.width = "width=" + width;
+        this.setModelObject(this.width);
```
martin-g: Why it is kept both as member field and as model object?

dmbeer:
@martin-g update to reflect suggestions. The widht is now consistent with the others. The reason for setting independately means that value can be retrived if they want them. I am happy to change this though to just pass the value. Not having a member as such just specific methods.

martin-g:
@dmbeer I've made some extra changes. Please review and tell me if you don't like it.
Thanks for the contribution!

dmbeer:
@martin-g I can see why it was changed to WebComponent, but changing userScale to boolean just adds more processing and conversion. Also if going to have a builder process I would move it a separate method.

martin-g:
'user-scalable' could have just two values: `yes` and `no`. So a boolean looks much better to me. Otherwise the application developer could set some invalid value
The processing time should be neglectable.

PR #592, opened 2016-03-26
Update Carousel.java

subes:
- fix page refresh not to hide the carousel since the active flag is missing
- make newImageList protected so users can customize item reuse settings etc


martin-g:
Please check https://github.com/l0rdn1kk0n/wicket-bootstrap/commit/5e396a0ea8fe657ff2191cbea4f7f709210f2c66. I think this simplifies the code and preserves the old behavior.

subes:
Yes, this is indeed much simpler.

PR #602, opened 2016-05-08
bootstrap based radio groups

reiern70:
just for review: ajax work against wicket 6.x but not against 7.3.0

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/radio/AjaxBooleanRadioGroup.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/radio/AjaxBooleanRadioGroup.java
@@ -1,0 +1,33 @@
+package de.agilecoders.wicket.core.markup.html.bootstrap.form.radio;
+
+import java.util.Arrays;
+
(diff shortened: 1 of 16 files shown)


martin-g:
It this ready to merge or it contains a quickstart inside showing the problem with BUBBLE ?

It is OK if I merge the radio group support from WB 0.9.x ? It will be easier for me to keep the git history this way.

martin-g:
I've merged WB master to wicket-7.x and the buttons seem to work.
Please report if you face any issues!
Thanks again!

reiern70:
Yes I fixed it 

https://github.com/reiern70/wicket-bootstrap/blob/668a5760ab5802edae2b75e8ab3c6e4a03840012/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/radio/BootstrapRadioGroup.java#L107

This is different from 6.x branch. That branch should also work.

There was a comment by @tahitiangabriel  about Bootstrap link. I will improve fix that in a different PR. 

Thanks to you for all the support!

I will try to create PRs for other components I use on my project that might be good to have on wicket-bootstrap.

martin-g:
OK. I've added this method override to 7.x too.

PR #619, opened 2016-05-31
Adding support for Bootstrap panels

heryqk:
http://getbootstrap.com/components/#panels

Added mods requested by @martin-g

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
@@ -27,12 +27,12 @@ import de.agilecoders.wicket.core.util.Components;
  * @param <T> the type of the panel's model object
  */
 public class BootstrapGenericPanel<T> extends GenericPanel<T>{
-	
+		
 	private static final String _PANEL_TITLE_ID = "panelTitle";
 	private static final String _PANEL_BODY_ID = "panelBody";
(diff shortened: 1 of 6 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
```diff
@@ -0,0 +1,171 @@
...
+		Panel panelFooter = newFooterPanel(_PANEL_FOOTER_ID, getModel());
+		add(panelFooter);
+		Components.hideIfModelIsEmpty(panelFooter);
+		
+	}
+	
+	@Override
+	protected void onBeforeRender() {
+	
+		add(new PanelBehavior(getPanelType()));
```
martin-g: What is the reason to re-add a new instance of the behavior on each render ?

martin-g:
Thanks!
Looks good! Just one question above about the re-adding of the PanelBehavior on each render of the panel.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
heryqk: It was a failed attempt at allowing the PanelType to be changed at runtime through an Ajax call for example, where through the click of a button, one could change the PanelType. 

I have made modifications. Submitting PR.

Thanks for your feedback.

martin-g:
Thank you!

Issue #639, opened 2016-10-03
NavBar Components with Position.RIGHT do not show up unless there is an item at Position.LEFT


dmbeer:
When adding NavBar Components ones set to the RIGHT do not show up unless there is and item to LEFT. 

navbar.addComponents(NavbarComponents.transform(Navbar.ComponentPosition.RIGHT,
                new NavbarButton<Void>(HomePage.class, Model.of("Overview")).setIconType(GlyphIconType.home)));

Does not show up unless I place an item before it to the LEFT.

navbar.addComponents(NavbarComponents.transform(Navbar.ComponentPosition.LEFT,
                new NavbarButton<Void>(HomePage.class, Model.of("Overview")).setIconType(GlyphIconType.home)));

Example Project can be found here https://github.com/dmbeer/wicket-bootstrap-example

PR #640, opened 2016-10-04
Fixed Navbar Not displaying components to the right properly - Issue #639

dmbeer:
Fixes Issue #639


martin-g:
Thank you, @dmbeer !

PR #641, opened 2016-10-21
Fix Bloodhound.Remote JSON serialization

iref:
I found bug in serialization of complex `typeaheadV10.bloodhound.Remote` configuration in TypeaheadV10 component.
Serializer uses `AbstractConfig.toJsonString` first and then writes it to output as string.
So configuration is serialized as string, not as JSON object and fetching from server doesn't work, because typeahead uses whole string as url.

Fixed by writing result of `AbstractConfig.toJsonString` as raw value.


martin-g:
Thank you!

Issue #646, opened 2016-11-18
Opinion? Overridable header close button markup for BootstrapModal


mikey75:
Hi,

What do you say to an idea of having the close button in the header overridable. For now the only way to make it something other than:

`<button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>`

is to create own markup for the whole modal. But suppose you want to have a button with an image and perhaps some custom ajax event on it. There's no way to do it without copying whole modal markup to your own instead of just extending the body part.

The idea od having most of the modal parts overridable (buttons, header label etc) is very cool but why the damn close button in the header has no wicket:id and the factory method? There is however an overridable createHeader() method but I don't get the idea of what it is for, since you cannot create a whole new header (both label and button) with it. Or maybe I can't see how. (hint me) 

Anyway what do you think? 
M.

martin-g:
Hint: send us a Pull Request! :-)
Thank you in advance!

mikey75:
I meant to, but first I need to know if that functionality is ever needed to anyone but me :)

mikey75:
https://github.com/l0rdn1kk0n/wicket-bootstrap/pull/647

mikey75:
Hello @martin-g :) bump. Any comments?

PR #647, opened 2016-11-19
issue #646

mikey75:
Override button

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Modal.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Modal.java
@@ -2,6 +2,8 @@ package de.agilecoders.wicket.core.markup.html.bootstrap.dialog;
 
 import de.agilecoders.wicket.core.markup.html.bootstrap.behavior.BootstrapResourcesBehavior;
 import de.agilecoders.wicket.core.markup.html.bootstrap.behavior.ICssClassNameProvider;
+import de.agilecoders.wicket.core.markup.html.bootstrap.button.BootstrapButton;
+import de.agilecoders.wicket.core.markup.html.bootstrap.button.Buttons;
 import de.agilecoders.wicket.core.util.Attributes;
 
 import de.agilecoders.wicket.jquery.util.Strings2;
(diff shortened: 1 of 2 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Modal.java
```diff
@@ -130,8 +135,24 @@ protected void populateItem(ListItem<Component> item) {
...
+     * 
+     * @param id The component id
+     * @return the close button
+     */
+    
+    protected Component createHeaderCloseButton(String id) {
+
+    	BootstrapButton btn = new BootstrapButton(id, Buttons.Type.Link);
+    	btn.setLabel(new  Model<String>("\u00D7"));
+    	btn.add(new AttributeModifier("class","close"));
```
martin-g: There is no need of this AttributeModifier. The class is already in the markup. Same for the label model.

mikey75: I wanted to retain the original markup. The button alone will change the markup. There's also css at play. If you don't override the button the modal will have no correct markup. I 'll try to post a quickstart when i come home.
cheers

mikey75: And, additionaly - actually - I think the better way would be to make header a container that would take components - same as  footer gets buttons etc. Suppose the getHeader factory method was meant for that.

martin-g: I'll take a look and test it tonight after work. No need of a quickstart. I can use the samples application.

mikey75: Any feedback?

martin-g:
I've simplified your version to use `WebMarkupContainer` instead of BootstrapButton and then there is no need to set the label and the CSS class in the Java code.

martin-g:
Thanks for the contribution!

mikey75:
Thank you!

PR #648, opened 2016-12-01
Change in datepicker's value triggers 'change' event of the field

vrozkovec:
Firing standard **change** event on value change. Otherwise only **dp.change** event is fired.

### DIFF
--- a/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePicker.java
+++ b/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePicker.java
@@ -7,6 +7,8 @@ import org.apache.wicket.model.IModel;
 
 import java.util.Date;
 
+import de.agilecoders.wicket.jquery.util.Strings2;
+
 /**
  * DateTime input field with a Eonasdan datetimepicker plugin.
  *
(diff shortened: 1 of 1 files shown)


bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePicker.java
```diff
@@ -60,6 +62,13 @@ public DatetimePicker(String id, IModel<Date> model, DatetimePickerConfig config
         this(id, model, config.getFormat());
         this.config = config;
     }
+    
+    @Override
+    public void renderHead(IHeaderResponse response) {
+    	super.renderHead(response);
+    	String scriptTemplate = "$('#%1$s').on('dp.change', function(){ $('#%1$s').trigger('change'); })";
+    	response.render(OnDomReadyHeaderItem.forScript(String.format(scriptTemplate, getMarkupId())));
```
martin-g: Use `Strings2.getMarkupId(component)` instead of `getMarkupId()`

martin-g:
Thanks!
I guess you need this change in Wicket 6/7 too, not only in 8.x ?
I'll cherry-pick it. Please just confirm!

vrozkovec:
Cool, thanks.  I still have some legacy projects using Wicket 6/7, but I plan to migrate all of them and I even haven't been using this component there.. so it is up to you, me personally I don't really need it.

martin-g:
OK. Then I won't downport it for now.

PR #654, opened 2016-12-22
Ajax datefield

reiern70:
No ready yet: I will add examples


martin-g:
Thanks, @reiern70 !
There is something wrong here. The PR is against `wicket-7.x` but the diff shows that Wicket 6.23.0 is specified in the pom.xml. It should be 7.5.0.

reiern70:
I will create a new PR.

PR #659, opened 2017-01-31
Added Font Awsome 4.7.0 CDN reference and icons

dmbeer:


martin-g:
The PR misses to update the version in the pom.xml: https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/wicket-7.x/bootstrap-extensions/pom.xml#L63

dmbeer:
Updated pom as suggested

martin-g:
Thanks!

PR #660, opened 2017-02-24
Czech localization of the BootstrapFileInputField

vrozkovec:
What should I do to propagate this to other branches? Should I create the file in each branch and make a pull request there? Won't there be a problem with branch merging then?

### DIFF
--- a/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/fileinput/BootstrapFileInputField_cs.properties
+++ b/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/fileinput/BootstrapFileInputField_cs.properties
@@ -1,0 +1,6 @@
+browseLabel=Procházet &hellip;
+removeLabel=Odebrat
+uploadLabel=Nahrát
+msgLoading=Nahrává se &hellip;
+msgProgress=Nahráno {percent}% z {file}
+msgSelected={n} souborů vybráno


martin-g:
IMO it would be better to use the XML variant of resource bundle because it works better for non-ASCII characters. See https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/Application_bg.properties.xml for example.

martin-g:
> What should I do to propagate this to other branches? Should I create the file in each branch and make a pull request there? Won't there be a problem with branch merging then?

Since your change is in `wicket-8.x` branch I will `cherry-pick` it in `master` and `wicket-7.x`. If it was in `master` then I would use `git merge`.

vrozkovec:
Now I see it is not really tied to the component, as res/fileinput.tmpl.js here makes no use of those variables:
```
PackageTextTemplate tmpl = new PackageTextTemplate(BootstrapFileInputField.class, "res/fileinput.tmpl.js");
Map<String, Object> variables = new HashMap<>();
variables.put("markupId", Strings2.getMarkupId(this));
for (String label : LABELS) {
    variables.put(label, getString(label));
}
variables.put("eventName", ajaxUploadBehavior.getEvent());

ajaxUpload = tmpl.asString(variables);

```

vrozkovec:
I can close this pull request and commit to the master then, if that's easier

vrozkovec:
Ok, I'll change it to the .xml version and commit to the master

vrozkovec:
Done here: #661 
closing this pull request

Issue #663, opened 2017-03-03
brandImage is stateful if no image is defined


ursjoss:
If you don't set a resource reference for the brandImage, the component is stateful. This prevents e.g. using a stateless form in a login page with the navbar on the same page. For any navbar with no visible brandImage, the component should be stateless IMHO.

ursjoss:
See https://github.com/l0rdn1kk0n/wicket-bootstrap/pull/664

martin-g:
#664 has been merged!

PR #664, opened 2017-03-03
Bug 663: brandImage should be stateless if not set

ursjoss:
https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/663


martin-g:
Thank you!

ursjoss:
@martin-g Thanks Martin. Out of curiosity - and for a potential next time:

* Should I have rather worked on master instead of the wicket-7.x branch?
* Does it make sense that I opened a bug report besides issueing a pull request?

Thanks and cheers
Urs

martin-g:
> Should I have rather worked on master instead of the wicket-7.x branch?

It's the same for me.


> Does it make sense that I opened a bug report besides issueing a pull request?

There is no need of an issue. PR is just enough. If something is not clear we can discuss in the PR itself.

ursjoss:
Thanks for the clarifications

PR #668, opened 2017-03-18
CheckBoxX: Timing issue with loading jquery before checkbox-x

ursjoss:
Good morning

I ran into some odd behavior with `CheckBoxX` on a wicket page together with `BootstrapMultiSelect`. In some cases, BootstrapMultiSelect did not render correctly in all circumstances. It boiled down to a timing issue. `jquery.js` might or might not have loaded before `checkbox-x.js`, depending on what else is going on rendering on the page. E.g. in the real project, it works when running with spring boot tooling in STS (including developer debug stuff etc.), but not when running the jar using java directlly...

I have created sample project removing most of the overhead out of the actual project to illustrate the issue:

https://github.com/ursjoss/sipamato-issue.git

The README of that project contains some screen shots, capturing the chrome developer console in a working and a failing page.

With the snapshot version of wicket-bootstrap-0.10.12-SNAPSHOT including this commit, all pages in the sample project load successfully.


bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/checkboxx/CheckBoxX.java
```diff
@@ -128,6 +130,7 @@ public CheckBoxXConfig getConfig() {
     public void renderHead(IHeaderResponse response) {
         super.renderHead(response);
 
+        response.render(JavaScriptHeaderItem.forReference(Application.get().getJavaScriptLibrarySettings().getJQueryReference()));
```
martin-g: This change could be simplified to:
response.render(JavaScriptHeaderItem.forReference(new `JQueryPluginResourceReference`(CheckBoxX.class, "js/checkbox-x.js")));

ursjoss: Totally, thanks Martin

martin-g:
Thank you for the contribution, @ursjoss !

ursjoss:
@martin-g Thanks for the merge - and the release :+1:

PR #671, opened 2017-05-09
Bootstrap select actions box config

vrozkovec:
Now it is possible to set Actions box on and off.
This config option makes sense only with BootstrapMultiSelect, but BootstrapSelectConfig is used also by BootstrapMultiSelect where these new config options make no sense. I've stated it in the comments of setters.
If needed, I can check whether getMultiple() returns true and set actions box only then or introduce new BootstrapMultiSelectConfig that inherits from BootstrapSelectConfig and contains only those new additions.


martin-g:
Thank you!

PR #682, opened 2017-06-08
added button block behavior setter

vrozkovec:


martin-g:
Thank you, @vrozkovec !

Issue #700, opened 2017-10-20
Material icons


tahitiangabriel:
Hi,

I would like to use Material icons from https://material.io/icons/

While glyphicon and fontawesome use the same kind of syntax (icon name is in the class) : 
`<i class="fa fa-address-book"></i>`

Material uses this syntax (the name of the icon is not in the class but between the tags):
`<i class="material-icons">3d_rotation</i>`

I do not see an easy way to provide a MaterialIconType that would works with the actual code.
I guess I would have to refactor Icon, IconBehavior and IconType.

@martin-g Any tought on this?

At least tomorrow I will provide an IIcon interface so one can provide its own implementation in the newIcon() function of the BootstrapButton.

martin-g:
Yes, it seems the current Icon impl won't be able to render such markup.
Some refactoring is required!

martin-g:
PR is merged!

PR #701, opened 2017-10-20
Refactor Icon and IconType to be able to use Material icons

tahitiangabriel:
PR for #700

Glyphicon and fontawesome syntax
`<i class="fa fa-address-book"></i>`

Material icons syntax
`<i class="material-icons">3d_rotation</i>

So I'have added a "tag body" field in the IconBehavior.
In the Icon class I do like a Label : I override `onComponentTagBody` if the tag body is provided.

An exemple of the MaterialIconType I've done :
```java
public class MaterialIconType extends IconType {
    /** serialVersionUID. */
    private static final long serialVersionUID = 1L;
    /** Material icons class. */
    private static final String MATERIAL_ICONS = "material-icons";

    public static final MaterialIconType threed_rotation = new MaterialIconType(MATERIAL_ICONS, "3d_rotation");
    public static final MaterialIconType input = new MaterialIconType(MATERIAL_ICONS, "input");
    public static final MaterialIconType dashboard = new MaterialIconType(MATERIAL_ICONS, "dashboard");

    /**
     * @param paramCssClassName the css classname
     * @param paramTagContent the tag body
     */
    protected MaterialIconType(final String cssClassName, final String tagBody) {
        super(cssClassName, tagBody);
    }

    @Override
    public String cssClassName() {
        return getCssClassName();
    }

}
```

I've also updated the javadoc.

Let me know if i need to tweak some more codes.

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/image/Icon.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/image/Icon.java
@@ -1,8 +1,12 @@
 package de.agilecoders.wicket.core.markup.html.bootstrap.image;
 
+import org.apache.wicket.markup.ComponentTag;
+import org.apache.wicket.markup.MarkupStream;
 import org.apache.wicket.markup.html.WebMarkupContainer;
 import org.apache.wicket.model.IModel;
 import org.apache.wicket.model.Model;
+
+import de.agilecoders.wicket.core.util.Components;
 
 /**
  * An Icon component displays a non localizable image resource.
@@ -11,7 +15,9 @@ import org.apache.wicket.model.Model;
  * @version 1.0
  */
 public class Icon extends WebMarkupContainer {
-
+    /** serialVersionUID. */
+    private static final long serialVersionUID = 1L;
+    /** Icon behavior. */
     private final IconBehavior iconBehavior;
 
     /**
@@ -24,14 +30,25 @@ public class Icon extends WebMarkupContainer {
         add(iconBehavior = new IconBehavior(type));
     }
 
-    public Icon(final IModel<IconType> type) {
-        this("icon", type);
+    /**
+     * @param typeModel the model containing the icon type
+     */
+    public Icon(final IModel<IconType> typeModel) {
+        this("icon", typeModel);
     }
 
+    /**
+     * @param id the component id
+     * @param type the icon type
+     */
     public Icon(final String id, final IconType type) {
         this(id, Model.of(type));
     }
 
+    /**
+     * 
+     * @param type the icon type
+     */
     public Icon(final IconType type) {
         this("icon", Model.of(type));
     }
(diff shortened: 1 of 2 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/image/Icon.java
```diff
@@ -66,7 +83,20 @@ public IconType getType() {
      * @return current icon type
      */
     public IconType type() {
-        return iconBehavior.type();
+        return getType();
     }
 
+    @Override
+    public void onComponentTagBody(MarkupStream markupStream, ComponentTag openTag) {
+        super.onComponentTagBody(markupStream, openTag);
```
martin-g: I think ` super.onComponentTagBody(markupStream, openTag);` should be in the `else` clause

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/image/Icon.java
```diff
@@ -66,7 +83,20 @@ public IconType getType() {
...
+        if(hasIconType() && getType().getTagBody() != null) {
+            replaceComponentTagBody(markupStream, openTag, getType().getTagBody());
+        }
+    }
+
+    @Override
+    protected void onComponentTag(ComponentTag tag) {
+        super.onComponentTag(tag);
+        Components.assertTag(this, tag, "i", "span");
```
martin-g: This might break someone's application.
I know that usually font icons are used with `i` or `span` but I've seen also with `em` and who knows what else could be in use in the wild.
Let's remove this check!

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/image/IconType.java
```diff
@@ -1,37 +1,64 @@
...
-	protected String getCssClassName() {
-		return cssClassName;
-	}
-    
-    @Override
-    public CssClassNameAppender newCssClassNameModifier() {
-        return new CssClassNameAppender(getCssClassName());
-    }
-}
+package de.agilecoders.wicket.core.markup.html.bootstrap.image;
```
martin-g: Now I see why I wasn't able to comment on your other PR - because I needed to check it with "No whitespace" option enabled.
Such full rewrite of the file due to different whitespace should be avoided.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/image/Icon.java
tahitiangabriel: You are right. Will change it.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/image/Icon.java
tahitiangabriel: In fact this is already done in the IconBehavior, so I can remove it!
So your `em` are already change in `i` with the actual code.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/image/IconType.java
tahitiangabriel: I don't see why all the whitespaces were changed. The actual code is using spaces and so am I in eclipse.
I have not this problem with diff in tortoisegit but I do see it in eclipse. What did I miss?

tahitiangabriel: It seems to have something something to do with linefeed and not space/tab.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=342372#c154

martin-g: Yes, the problem is that your Eclipse have changed the newlines to Windows style.

martin-g:
Thanks, @tahitiangabriel !

Issue #699, opened 2017-09-22
For components that call instance methods during initialization, move initialization from the constructor to onInitialize()


ageery:
Here's a concrete example. 

I would like to create a non-abstract sub-class of DropDownButton which, rather than having to sub-class DropDownButton to provide the list of buttons, would pass a lambda in the constructor to provide the list of buttons.

However, I can't do this because the constructor of DropDownButton calls newSubMenuButtons. The DropDownButton constructor is called before the constructor for my sub-class is called. This means that my sub-class, which implements the newSubMenuButtons method, can't see the lambda that has been passed in via the constructor because the sub-class's constructor hasn't yet been executed which sets an instance variable with the value of the constructor parameter.

The solution to this is simply to move the logic from the constructor and put it into the Wicket onInitialize() method. Everything should work exactly the same; the only difference is that the component is a little more flexible for sub-classing.

Does that make sense? I can send some pull requests of components I'm interested in changing (another example is Modal).

See also: https://stackoverflow.com/questions/18348797/why-is-it-considered-bad-practice-to-call-a-method-from-within-a-constructor

Thank you!

martin-g:
PR is welcome!

martin-g:
Closing since there is no interest in this improvement.
Please send a PR if you still need it!

PR #702, opened 2017-10-20
button/link : be able to not use the splitter

tahitiangabriel:
I've got a use case where I don't want the splitter (&nbsp;) between the icon and the label (my label contains html code) in the buttons or links.
For exemple when I put HTML code in the label.

So I've added a `useSplitter(boolean value)` in all the Button and Link.

I've also fix the `setLabel(IModel<String> label)` not updating the button's model as the model is store in
the label and also in the button.

The BootstrapAjaxFallbackButton class is now similar as the other buttons/links (with newLabel, newIcon, newSplitter functions).

how to disable the splitter : 
```java
new BootstrapLink<Void>("dashboardLink", Type.Menu) {
            public void onClick() {
                // TODO task
            }
        }.useSplitter(false).setIconType(FontAwesomeIconType.dashboard).setLabel(Model.of("Dashboard"))
```

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/BootstrapAjaxButton.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/BootstrapAjaxButton.java
@@ -26,6 +26,8 @@ public abstract class BootstrapAjaxButton extends AjaxButton implements IBootstr
     private final Component label;
     private final Component splitter;
     private final ButtonBehavior buttonBehavior;
+    /** To use the splitter or not (true by default). */
+    private boolean useSplitter = true;
 
     /**
      * Construct.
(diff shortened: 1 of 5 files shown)


martin-g:
For some reason (GitHub bug ?!) I cannot comment inline in the PR, i.e. on specific lines...

martin-g:
If you need to introduce the factory methods (like  `newIcon()`) then please move the code from the constructor to `onInitialize()`.

tahitiangabriel:
I known that calling overridable methods in the constructor is a bad practice but that is how it is done in all those classes.
Should I change all of them? And so store the label's model and the button's type as class properties?

martin-g:
Let's leave this improvement to https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/699

tahitiangabriel:
@martin-g There use to be a snapshot available at Sonatype. It does not seem to be up to date.
Is there another place where one can get the snapshot?

martin-g:
Manually deployed them to Sonatype!

Issue #708, opened 2017-11-08
Bootstrap Wizard


vrozkovec:
Hi, would you be interested in Wizard component in Bootstrap style? 
For view I'd use Bootstrap's panels and for links Bootstrap buttons.

martin-g:
Sure!

You mean Wicket Extensions' Wizard component, right ? Just styled with Bootstrap.
Or completely new one ?

vrozkovec:
The one from the Extensions.

martin-g:
Implemented with #712

PR #712, opened 2017-11-29
Added Bootstrap styled Wizard component

vrozkovec:
See issue #708


martin-g:
Awesome! Thanks, @vrozkovec !

It can become even more awesome if you add an example in the samples! :-)

PR #727, opened 2018-03-12
NotificationPanel fix for not showing rendered messages.

### LINKED ISSUE #723: NotificationPanel does not respect value set by showRenderedMessages(boolean showRenderedMessages)
If you call `showRenderedMessages(final boolean showRenderedMessages)` on notification panel, value is set, but too late. 

This code
```
@Override
    protected FeedbackMessagesModel newFeedbackMessagesModel() {
        FeedbackMessagesModel model = super.newFeedbackMessagesModel();

        if (!showRenderedMessages) {
            model.setFilter(new IFeedbackMessageFilter() {
                @Override
                public boolean accept(FeedbackMessage message) {
                    return !message.isRendered();
                }
            });
        }

        return model;
    }
```
is called from the constructor meaning value of `showRenderedMessages` is always false.

vrozkovec:
newFeedbackMessagesModel() is called from the constructor
and because of that calling
showRenderedMessages(boolean) later had no effect

Fixes #723

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/common/NotificationPanel.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/common/NotificationPanel.java
@@ -18,7 +18,7 @@ import org.apache.wicket.util.time.Duration;
 public class NotificationPanel extends FencedFeedbackPanel {
 
     private Duration duration;
-	private IFeedbackMessageFilter SKIP_RENDERED_MESSAGES_FILTER = new IFeedbackMessageFilter()
+	private static final IFeedbackMessageFilter SKIP_RENDERED_MESSAGES_FILTER = new IFeedbackMessageFilter()
 	{
 		
 		@Override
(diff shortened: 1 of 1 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/common/NotificationPanel.java
```diff
@@ -18,8 +18,16 @@
 public class NotificationPanel extends FencedFeedbackPanel {
 
     private Duration duration;
-    private boolean showRenderedMessages = false;
-
+	private IFeedbackMessageFilter SKIP_RENDERED_MESSAGES_FILTER = new IFeedbackMessageFilter()
```
martin-g: Please add `static final`

vrozkovec: added

martin-g:
Thank you, @vrozkovec !

martin-g:
Cherry picked it to 7.x and 8.x branches!

PR #738, opened 2018-04-04
fix for checkbox' label model

vrozkovec:
Passing empty model for checkbox label for the case of using the BootstrapToggle in the component with CompoundPropertyModel where component tries to find a model for a label and tries that with a "post-label" property expression on a CPM's object.


martin-g:
Thank you, @vrozkovec !

PR #740, opened 2018-04-27
Wicket 8.x - BootstrapLabel component and text and choice filter property columns

namtar:
Added a Bootstrap Label Component (https://getbootstrap.com/docs/3.3/components/#labels). Its implementation is done analog to the BootstrapButton´s implementation. 
Added a BootstrapTextFilteredPropertyColumn and a BootstrapChoiceFilteredPropertyColumn to be used with the BootstrapDefaultDataTable. I Also added a filtered datatable example linked in a separate "example" nav dropdown for the demo application. If this example is not wanted in the demo application I will remove it. I thought it would not be right placed in the "BaseCSS" or "Components" section so I added an example navigation dropdown.

The Datatable example looks like this.
datatableexamplepage


martin-g:
Thank you for the contribution	!

PR #746, opened 2018-06-25
Upgrade core components to bootstrap 4

iref:
I updated core components to bootstrap 4. Changes follow [Bootstrap Migration Guide](https://getbootstrap.com/docs/4.1/migration/). Most of the changes are in css classes and markup, but there are some breaking API changes. These are caused by components no longer beeing supported in v4 (labels, wells, and so on) or by bootstrap 4 renaming some modifiers that are in our enums, e.g. `.text-default` context class is now `.text-secondary`.


martin-g:
Thank you, @iref !

ffcsoftware:
Is there any reason why the optional internal container (to avoid full width) for the NavBar component was removed in this commit? I see that for v4 is still a recommended practice to limit width [https://getbootstrap.com/docs/4.0/components/navbar/#containers](https://getbootstrap.com/docs/4.0/components/navbar/#containers)

I'm not at expert at all at Bootstrap, maybe I'm missing something.

PS Thank you all for your hard work.

martin-g:
@ffcsoftware Please create a new Pull Request if you need this feature! Thank you!

PR #750, opened 2018-07-25
Bootstrap4 extensions

iref:
I updated extensions to support bootstrap 4. Unfortunately, not all are available at the moment.

## Unsupported extensions

* jasny - https://github.com/jasny/bootstrap/issues/438
   InputMask was changed for https://github.com/RobinHerbots/Inputmask
* dropdown auto open - Not developed for 3 years (bootstrap-hover-dropdown)
* Context Menu - Not developed for 2 years (https://github.com/dgoguerra/bootstrap-menu)
* Bootstrap Select
* X-Editable
* Bootstrap Slider - https://github.com/seiyria/bootstrap-slider/issues/689

## Extension Changes

* Tempus Dominus datetimepicker (updated version of eions datetimepicker)

PR #751, opened 2018-07-25
Remove slider page from sample app

iref:
Forgot to remove `SliderPage` from sample after slider was removed in #750, which caused compilation errors.


reiern70:
No version os slider for 4?

iref:
Unfortunately, not yet. There is opened issue, though. https://github.com/seiyria/bootstrap-slider/issues/689

PR #752, opened 2018-08-22
Popper.js customizable sesttings

iref:
I run into bug when bootstrap uses CDN to load resources instead of webjars. New dependency on popper.js didn't load from CDN but webjars and it caused application to fail on startup.

I added CDN setup for popper.js to fix this.


martin-g:
Thank you, Jan!

PR #754, opened 2018-08-27
Add a note about stripping wicket tags to the README

### LINKED ISSUE #753: [bs4] Navbar issue with wicket:panel tag
Versions: `wicket-bootstrap:3.0.0-M1` (with `spring-boot:2.0.4.RELEASE` and `wicket-spring-boot:2.1.1`)

When `MarkupSettings#stripWicketTags==false` the `<wicket:panel>` tag within the `<nav>` tag which holds the navbar components seems to break some css rules to the extend, that the navbar components are rendered below the navbar brand and not right next to it.

Stripping the wicket tags fixes the "issue". Unsure if this qualifies as a bug and even more so if there's an easy fix for this, if it needs fixing at all. Just wanted to mention.

mweirauch:
Resolves #753

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,11 @@ Bootstrap.install(this, settings);
 
 then you are able to use all wicket-bootstrap components.
 
+Noteworthy
+----------
+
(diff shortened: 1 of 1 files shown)


martin-g:
Thank you, @mweirauch !

PR #762, opened 2018-10-26
Update DropDownButton.html

helmut8080:
Adapt to bootstrap 4. (ul, li removed, new css classes)

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
@@ -4,15 +4,16 @@
     </head>
     <body>
         <wicket:panel>
-            <a class="btn dropdown-toggle" data-toggle="dropdown" wicket:id="btn" href="#">
-                <span wicket:id="icon"></span>
-                <wicket:container wicket:id="label"></wicket:container>
-            </a>
-            <ul wicket:id="dropdown-menu" class="dropdown-menu">
-                <li wicket:id="buttons">
-                    <a wicket:id="button" class="dropdown-item"></a>
-                </li>
-            </ul>
+           <a class="dropdown-toggle" wicket:id="btn" href="#" type="button" 
+		  		data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+		  		<span wicket:id="icon"></span>
+		    	<wicket:container wicket:id="label"></wicket:container>
+		  </a>
+		  <div wicket:id="dropdown-menu" class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+		  	<div wicket:id="buttons">
+		    	<a wicket:id="button" class="dropdown-item" href="#">Action</a>
+		    </div>
+		  </div>
         </wicket:panel>
     </body>
 </html>
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
@@ -4,14 +4,14 @@
     </head>
     <body>
         <wicket:panel>
-           <a class="dropdown-toggle" wicket:id="btn" href="#" type="button" 
-		  		data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+           <a class="dropdown-toggle" wicket:id="btn" type="button" 
+		  		data-toggle="dropdown" aria-haspopup="true" href="javascript:;" aria-expanded="false">
 		  		<span wicket:id="icon"></span>
 		    	<wicket:container wicket:id="label"></wicket:container>
 		  </a>
 		  <div wicket:id="dropdown-menu" class="dropdown-menu" aria-labelledby="dropdownMenuLink">
 		  	<div wicket:id="buttons">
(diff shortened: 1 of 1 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
```diff
@@ -4,15 +4,16 @@
...
-            <a class="btn dropdown-toggle" data-toggle="dropdown" wicket:id="btn" href="#">
-                <span wicket:id="icon"></span>
-                <wicket:container wicket:id="label"></wicket:container>
-            </a>
-            <ul wicket:id="dropdown-menu" class="dropdown-menu">
-                <li wicket:id="buttons">
-                    <a wicket:id="button" class="dropdown-item"></a>
-                </li>
-            </ul>
+           <a class="dropdown-toggle" wicket:id="btn" href="#" type="button"
```
martin-g: Please either remove `href="#"` or use `href="javascript:;"` instead. Using `"#"` has bad side effect of scrolling the page to the top.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
```diff
@@ -4,15 +4,16 @@
...
-                </li>
-            </ul>
+           <a class="dropdown-toggle" wicket:id="btn" href="#" type="button" 
+		  		data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+		  		<span wicket:id="icon"></span>
+		    	<wicket:container wicket:id="label"></wicket:container>
+		  </a>
+		  <div wicket:id="dropdown-menu" class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+		  	<div wicket:id="buttons">
+		    	<a wicket:id="button" class="dropdown-item" href="#">Action</a>
```
martin-g: Same for `href="#"`.

martin-g: Is the text `Action` only for a preview here ?

helmut8080: - href="#" exchanged against href="javascript:;" 
- Action removed (just needed for my project)

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
tahitiangabriel: Why not use the <button> tag instead ?
https://getbootstrap.com/docs/4.0/components/dropdowns/#single-button-dropdowns

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
```diff
@@ -4,15 +4,16 @@
...
-                    <a wicket:id="button" class="dropdown-item"></a>
-                </li>
-            </ul>
+           <a class="dropdown-toggle" wicket:id="btn" href="#" type="button" 
+		  		data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+		  		<span wicket:id="icon"></span>
+		    	<wicket:container wicket:id="label"></wicket:container>
+		  </a>
+		  <div wicket:id="dropdown-menu" class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+		  	<div wicket:id="buttons">
```
tahitiangabriel: I think we should use a wicket-container otherwise it will produce an extra unwanted div:

<div wicket:id="dropdown-menu" class="dropdown-menu">
                <wicket:container wicket:id="buttons">
                    <a wicket:id="button" class="dropdown-item" href="javascript:void(0)"></a>
                </wicket:container>
            </div>

helmut8080: ButtonList.populateItem(ListItem<AbstractLink> item) is overwritten with item.setRenderBodyOnly(true). So no extra unwanted div is created. 
However, using wicket:container is for sure the more elegant version. I will change to wicket:container

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
helmut8080: Because I want to enable DropUp, DropRight ... variants, However, did not try if this works with the button tag. Its just described with btn-grp.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/DropDownButton.html
tahitiangabriel: That's weird because I do have the extra div in my code at the end...

helmut8080: With item.setRenderBodyOnly(true) you should not. I did not have the additional div. However its now changed to wicket:container. So if you use the current DropDownButton class you will get the extra div.

PR #770, opened 2018-12-27
Update Navbar.html

x-paphio:
adding ml-auto will allow a design without anything in the left enclosure to work. Also adding align-items-center class to the ul tag make for a more streamline design, but this is personal taste.


martin-g:
Thank you, @x-paphio !

PR #773, opened 2019-02-05
Chore/dependencies

ursjoss:
This pull request bumps some of the dependencies that are not up to date. Please note that the project compiles and the tests run (see remark to wicket-webjars-2.0.10 below).

I tried to be cautious with the updates, not going all the way to the latest version available if the changelog indicated incompatible changes (see remark to summernote below).

I did some smoke tests in the samples module and made sure all tests in my application passed (having wicket-bootstrap as a dependency). I also did some smoke tests there. *However*: I presume some more testing is advised. Especially since I don't personally use all the dependencies that are updated in this PR.

A few are to be mentioned specifically:

* wicket-webjars-2.0.10: The test is pulled along to match the changes that were introduced in https://github.com/l0rdn1kk0n/wicket-webjars/pull/57. Doing some smoke tests both in the sample application as well as in my own application depending on wicket-bootstrap, I have the impression that the behavior is working correctly. However, I'm not sure if the current modification of the test is what you'd want to have in the project. Also, if and when https://github.com/l0rdn1kk0n/wicket-webjars/pull/58 is going to be merged, the test will most likely have to be adjusted to that change as well.

* summernote-0.8.7 bumps jquery 2.1.4 to 3.2.1 and summernote-0.8.10 even switches to bootstrap-4. So I stick with summernote-0.8.6

There are a few dependencies that I did not touch, namely:

* javax:javaee-web-api 7.0 -> 8.0+
* javax.servlet:javax.servlet-api 3.1.0 -> 4.0.1
* mockito-core 1.10.19 -> 2.x (2.24.0)
* org.eclipse.jetty.aggregate:jetty-all-server
* newrelic-api 3.30.0 -> 3.48.0 (or even 4.10.0)
* com.google.javascript:closure-compiler
* org.webjars:jquery
* org.webjars:bootstrap

Even though it seems not be used extensively, changing to mockito-core 2.x would probably make a good pull request by itself. I could give it a go if you're interested. Let me know.


martin-g:
Thanks for the PR, @ursjoss !

About the notes:
* I will check the latest state of wicket-webjars. I intentionally use the older version due to the problems
* we can use summernote-0.8.10 in wicket-8.x-bootstrap-4.x branch (released as 3.0.0-Mx)
* javax:javaee-web-api 7.0 -> 8.0+  - (should be the one that includes Servlet API 3.1.0)
* javax.servlet:javax.servlet-api 3.1.0 -> 4.0.1  (should use 3.1.0, as Wicket 8.x)
* mockito-core 1.10.19 -> 2.x (2.24.0)  (new PR is welcome! You can even replace JUnit with 5.4.0)
* org.eclipse.jetty.aggregate:jetty-all-server (this is OK to upgrade to the latest. It is used only in the sample app)
* newrelic-api 3.30.0 -> 3.48.0 (or even 4.10.0) (This one could be dropped completely, I think)
* com.google.javascript:closure-compiler  (should be OK to upgrade)
* org.webjars:jquery   (latest 2.x for `wicket-8.x`, latest 3.x for `wicket-8.x-bootstrap-4.x` branch)
* org.webjars:bootstrap -    (latest 3.x for `wicket-8.x`, latest 4.x for `wicket-8.x-bootstrap-4.x` branch)

martin-g:
The plan looks good!
About Jetty: I think we can upgrade to latest 9.4.x. AFAIR it is used only for `mvn jetty:run` and `StartSamples.java` so it should not affect the end users' applications.

ursjoss:
@martin-g I have completed the tasks we have discussed here.

bootstrap-samples/pom.xml
```diff
@@ -138,29 +143,17 @@
...
-                        </connector>
-                        <connector implementation="org.eclipse.jetty.server.ssl.SslSocketConnector">
-                            <port>8443</port>
-                            <maxIdleTime>3600000</maxIdleTime>
-                            <keystore>${project.build.directory}/test-classes/keystore</keystore>
-                            <password>wicket</password>
-                            <keyPassword>wicket</keyPassword>
-                        </connector>
-                    </connectors>
+                    <jettyXml>jetty.xml,jetty-http.xml,jetty-ssl.xml,jetty-https.xml</jettyXml>
```
martin-g: Let's move these files into `src/test/jetty` to not pollute the root folder of the module.

ursjoss: Sure, I'll do so tonight

ursjoss: Done

martin-g:
Thank you, @ursjoss !

PR #785, opened 2019-02-09
Report coverage of wicket-8.x branch to coveralls

ursjoss:


martin-g:
Thank you, @ursjoss !
But I expect not very pleasant results :-)

PR #790, opened 2019-02-09
Feature/707 font awesome 5

### LINKED ISSUE #707: Font Awesome 5
Hi, would you be interested in creating icon types for Font Awesome 5?
It is paid version, so including proper resources would be up to the user.

ursjoss:
I'm somewhat unsure what would be the best way to implement font-awesome-5.x.

This might eventually fix #707.

First thing, changing the version from 4.7 to 5.7 is a breaking change, as (at least some) of the icons have different names or are not available anymore.

Further, font-awesome now has 4 different styles of icons, 3 of which can be used in wicket-bootstrap.

* style `solid`: All icons are free and can be used in the project. To be used with `fas` (or `fa`)
* style `regular`: Only some of those icons are free, many are not. To be used with `far`
* style `brand`: all free. To be used with `fab`
* style `light` all of which are not free - and thus don't need to be further discussed here.

The icons of style `brand` have distinctive names and are not an issue. The icons that are freely available in style `regular` use the same name as those in `solid`.

I have separated the three different free icon styles within FontAwesomeIconTypeBuilder. One enum each `FontAwesomeSolid`, `FontAwesomeRegular` and `FontAwesomeBrand` (all implementing `FontAwesomeGraphic` which is now an interface) define each icon. So as with the additional attributes of the icons (spin, fixedWith etc.), the user has full freedom on how to use which style in which variant.

For the pre-built instances of `FontAwesomeIconType` I see two general approaches (with sub-options) how we can handle those in wicket-bootstrap:

* Aproach 1: Every style is implemented in it's own instance of `FontAwesomeIconType`. We can use tags to distinguish the style (between regular and solid, no name clash for brands)
  * Option 1.1: Use the tag `_r`/`_s` where we have both a variant of style solid and regular (e.g. `address_book_s` and `address_book_r`). Where we only have a solid variant right now, we don't apply the tag `s`. the icon `adjust` currently is only (freely) available as solid. If in the future it will also have a regular variant that is free, we would add `adjust_r` but would also need to rename `adjust` to `adjust_s` for consistency.
    * **Pro:** Some of the icons with only a solid variant will get the same name with font-awesome-5.7 as with font-awesome-4.7. Less migration required for the users of wicket-bootstrap with the version bump.
    * **Cons:** A new icon variant will break the name of the pre-existin solid variant.
  * Option 1.2: Use tags `_s` (solid) or `_r` (regular) with *all* icons of those style, regardless of whether we currently have a solid and a regular (free) variant or not. `address_book_s` and `address_book_r`, but also `adjust_s` (with no `adjust_r`).
   * **Pro:** Once an icon of style regular will become free in a later version of font-awesome, we will only have to add that regular variant `adjust_r` without having to bother about the previously existing solid variant, which is already called `adjust_s`.
   * **Cons**: *All* icons will have a different name with the version change compared to the icon names in wicket-boostrap with font-awesome 4.7.0.
 
I strongly prefer option 2 and have already implemented this pattern within this pull request. It will require more migration work initially for users of wicket-bootstrap, but only once. Later amendments will be simpler.

I was also thinking about a completely separate approach with enhancing the prebuilt instances of `FontAwesomeIconType` to handle both styles (regular, solid) if both are available. If not, we could throw exception if somebody wants to use a non-existing style of an icon. I have the feeling that this would be more aligned with font-awesomes approach with the separate styles. But I'm not sure if and how this could be implemented to play along nicely with the usage of IconType implementations. It would certainly be much more complex to implement. Not sure if it's worthwhile.

References:

* https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4
* https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use

Note: While the documentation states that regular style requires a pro license, the detail page about icons lists some icons of the regular style as pro (e.g. https://fontawesome.com/icons/ad?style=solid), some as not (https://fontawesome.com/icons/address-book?style=solid). I have only implemented the regular icons that are listed as free (https://fontawesome.com/icons?d=gallery&m=free).

What do you think?


martin-g:
At the weekend I've asked via mail:

```
For better backward compatibility we may have two sets of fonts/classes: one for 4.x and another for 5.x.
What do you think? 
```

but just now I received a `Mail Delivery failed` email.

ursjoss:
Hi @martin-g . Excellent idea. I will have a look at how to tackle that tonight or later this week.

ursjoss:
@martin-g Ok, I have pushed some commits.

* renamed all classes and types related with font-awesome 5.x to ...FontAwesome5...
* restored the old classes (with their original names)
* added font-awesome-4.7.0 as dependency in the pom
* Deprecated the 4.7.0 classes

I was considering renaming the old 4.7 classes to ...FontAwesome4... and keep the version agnostic names to the font-awesome-5.x classes. But that would not allow to keep the migration effort minimal now. Not sure what's better.

If you'd like to keep the deprecation, it would be nice to add the comment in the javadoc about when you intend to remove the old classes. Otherwise we can just drop the deprecation entirely.

Let me know how you'd like to proceed.

martin-g:
All sounds good!
No need of deprecation for 4.x.

ursjoss:
@martin-g ok, the commit that introduced the deprecation is reverted.

bootstrap-extensions/pom.xml
```diff
@@ -134,10 +139,6 @@
                     <groupId>org.webjars.bower</groupId>
                     <artifactId>bootstrap</artifactId>
                 </exclusion>
-                <exclusion>
```
martin-g, line 137: Are you sure about this ?
The managed dependency above (line 61) has different `groupId`. I have the feeling that now we may have two kind of dependencies to `font-awesome`.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/icon/FontAwesome5CssReference.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    public static FontAwesome5CssReference instance() {
+        return Holder.INSTANCE;
+    }
+
+    /**
+     * Private constructor.
+     */
+    private FontAwesome5CssReference() {
+        super("font-awesome/current/css/all.css");
```
martin-g, line 30: How does this figures out to use v5 and not v4 ?
Is v4 in a different folder ? E.g. `font-awesome4/...` ?

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/icon/FontAwesome5IconType.java
```diff
@@ -0,0 +1,1268 @@
...
+import de.agilecoders.wicket.extensions.markup.html.bootstrap.icon.FontAwesome5IconTypeBuilder.FontAwesome5Regular;
+import de.agilecoders.wicket.extensions.markup.html.bootstrap.icon.FontAwesome5IconTypeBuilder.FontAwesome5Solid;
+import org.apache.wicket.util.string.Strings;
+
+import static de.agilecoders.wicket.extensions.markup.html.bootstrap.icon.FontAwesome5IconTypeBuilder.on;
+
+/**
+ * All the icon types.
+ */
+public class FontAwesome5IconType extends IconType {
```
martin-g, line 14: I have always wondered how people create these classes ?
Did you create it manually ? Or with a script ?
I ask because I guess one day I will have to update it.

martin-g (approved):

bootstrap-extensions/pom.xml
ursjoss: Well summernote-0.8.x is not pulling in a transitive dependency on font-awesome anymore, so there's not use excluding it.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/icon/FontAwesome5CssReference.java
ursjoss: the css for v4 is in a file `font-awesome.css` while that for v5 is in `all.css`. But the problem is another one IMHO, I'll write that in a general comment below.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/icon/FontAwesome5IconType.java
ursjoss: It wasn't one of my favorite chores either :-) I copied all the icon names from the fontawesome page (https://fontawesome.com/icons?d=gallery&m=free filtering by style) and pasted that into a reasonable editor.

I then used regex replacement to get from 

```
ad
address_book
```

to

```
    public static final FontAwesome5IconType ad_s = on(FontAwesome5Solid.ad).build();
    public static final FontAwesome5IconType address_book_r = on(FontAwesome5Regular.address_book).build();
```
and just pasted it into the class file I had copied from the v4 version in the project.

bootstrap-extensions/pom.xml
martin-g: :+1:

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/icon/FontAwesome5IconType.java
martin-g: Regex FTW! :-)

martin-g:
Merged it!
If we find issues we will fix them with a new PR.
Thank you, @ursjoss !

ursjoss:
@martin-g Sorry, I was in a meeting in between commenting here. There is an issue I did not notice before: We cannot have a dependency on both versions in the pom, they have the same `groupid` and `artifactId` and cause maven a head-ache.

martin-g:
In that case I suggest to:

1. In branch `wicket-8.x`: 
* remove the dependency for 5.x from pom.xml
* add a javadoc to FontAwesome5 classes explaining that the application should provide the dependency, i.e. to upgrade to 5.x
2. In branch `wicket-8.x-bootstrap-4.x` we can give FA 5.x higher priority, i.e. make it available by default and the user can downgrade it in the application's pom.xml by specifying 4.x

The question is what to do with Summernote and Markdown in `wicket-8.x` ?

ursjoss:
@martin-g I won't be able to work on this until at least Sunday evening. I'll have a look asap.

martin-g:
I have done it earlier today.
Any ideas how to make it better are welcome!

ursjoss:
@martin-g Ok, thanks.

PR #791, opened 2019-02-15
Update jquery.wb.validation.js

x-paphio:
Without this changege the red square will stay on the field when the tooltip disapear.


martin-g:
Thank you, @x-paphio !

PR #798, opened 2019-04-17
Add new and missing icons up-to and including font-awesome-5.8.1

ursjoss:
Adding font-awesome-icons that were
* available but not defined with font-awsome-5.7.1
* are new with font-awesome-5.8.0 (see https://github.com/FortAwesome/Font-Awesome/blob/master/CHANGELOG.md)

Also changed the comments illustrating how to pull in font-awesome-5.8.1.


martin-g:
Thank you, @ursjoss !

PR #807, opened 2019-06-04
Explicitly add .navbar-left class for left aligned components

vrozkovec:
This class adds float:left !important to the components which may be needed for some templates.


martin-g:
Thank you, @vrozkovec !
Once Wicketstuff Core 8.5.0 is released I will release WB as well!

martin-g:
2.0.10 is released!

PR #810, opened 2019-09-09
Add onLabel and offLabel setters to BootstrapCheckBoxPickerConfig

vrozkovec:


vrozkovec:
Closing, missed those:
```
    protected IModel<String> getOffLabel() {
        return Model.of("No");
    }

    protected IModel<String> getOnLabel() {
        return Model.of("Yes");
    }
```

PR #819, opened 2019-10-08
Bump BootstrapCheckboxPicker to v2.0.0

vrozkovec:


martin-g:
There is a small API break related to the `style` / `size` config settings but I think it should be easy to migrate.
Thanks, @vrozkovec !

Issue #777, opened 2019-02-07
Upgrade bootstrap to 3.4.0 in `wicket-8.x` branch


ursjoss:
As discussed in #773

martin-g:
Done!

tgoetz:
I think the upgrade of the Modernizr lib from 2.8.3 to 2.8.3-1 introduced a bug:
* the webjars artifact for modernizr-2.8.3-1 only contains modernizr.min.js, but no modernizr.js
* in de.agilecoders.wicket.core.markup.html.references.ModernizrJavaScriptReference the name is set to "/modernizr/current/modernizr.js"
* in our application this leads to a MissingResourceException, because the modernizr resource can't be found:

Caused by: java.util.MissingResourceException: Unable to get cache key for [webjars js resource] /modernizr/current/modernizr.js (resolved name: /webjars/modernizr/null/modernizr.js)
	at org.apache.wicket.resource.bundles.ConcatBundleResource.reportError(ConcatBundleResource.java:263)
	at org.apache.wicket.resource.bundles.ConcatBundleResource.getCacheKey(ConcatBundleResource.java:238)
	at org.apache.wicket.request.resource.caching.version.CachingResourceVersion.getVersion(CachingResourceVersion.java:101)
	at org.apache.wicket.request.resource.caching.FilenameWithVersionResourceCachingStrategy.decorateUrl(FilenameWithVersionResourceCachingStrategy.java:105)
	at org.apache.wicket.core.request.mapper.BasicResourceReferenceMapper.mapHandler(BasicResourceReferenceMapper.java:234)
	at org.apache.wicket.request.mapper.ParentPathReferenceRewriter.mapHandler(ParentPathReferenceRewriter.java:92)
	at org.apache.wicket.request.mapper.CompoundRequestMapper.mapHandler(CompoundRequestMapper.java:210)
	at org.apache.wicket.request.cycle.RequestCycle.mapUrlFor(RequestCycle.java:451)
	at org.apache.wicket.request.cycle.RequestCycle.urlFor(RequestCycle.java:551)
	... 84 common frames omitted

tgoetz:
I debugged this and noticed that in de.agilecoders.wicket.webjars.collectors.AssetsMap#findRecentVersionFor there will be returned null, as the filename ('modernizr.js', set in ModernizrJavaScriptReference) does not match the real file name (modernizr.min.js) ...

tgoetz:
[...]

PR #822, opened 2019-12-06
Fixing incorrect resource name for modernizr.js (#777)

tgoetz:
See https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/777

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/references/ModernizrJavaScriptReference.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/references/ModernizrJavaScriptReference.java
@@ -31,7 +31,7 @@ public class ModernizrJavaScriptReference extends WebjarsJavaScriptResourceRefer
      * Private constructor.
      */
     private ModernizrJavaScriptReference() {
-        super("/modernizr/current/modernizr.js");
+        super("/modernizr/current/modernizr.min.js");
     }
 
 }


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/references/ModernizrJavaScriptReference.java
```diff
@@ -31,7 +31,7 @@ public static ModernizrJavaScriptReference instance() {
      * Private constructor.
      */
     private ModernizrJavaScriptReference() {
-        super("/modernizr/current/modernizr.js");
+        super("/modernizr/current/modernizr.min.js");
```
martin-g, line 34: Is this really needed ? Wicket should use the min version in PROD mode

tgoetz: I have not tested in PROD mode yet, but in DEV mode I get the mentioned exception ...

tgoetz: I'd say the real question is: why does the webjars artifact of modernizr-2.8.3-1 not contain modernizr.js? modernizr-2.8.3 contains both (modernizr.js and modernizr.min.js) ...

tgoetz: @martin-g what do you think? I currently created a workaround for this in my current project (created my own ModernizrJavaScriptReference with the currected path and configured a global resource replacement in Wicket).

martin-g:
I'd rather go back to the earlier version with both files.

tgoetz:
Ok, this is also a suitable solution I guess. I thought that upgrading modernizr to 2.8.3-1 might have been done for a reason ...?! But if you say it's safe to go back to 2.8.3 I will provide a pull request for this.

PR #826, opened 2019-12-22
Move panel's content initialization to the onInitialize method.

vrozkovec:
Also instead of setting null model, pass empty model - this way when calling `Components.hideIfModelIsEmpty` upon panel's initialization, there is no hierarchy traversal that tries to get model's object value with the `panelFooter` and `panelBody` property lookups.


martin-g:
Thanks, @vrozkovec !

PR #831, opened 2020-01-22
Fixed wrong path in minified css of the BootstrapFileInputField

vrozkovec:
`../img/loading.gif` --> `../res/loading.gif`


martin-g:
I think it would be better to relocate `loading.gif` in `img` folder.
This way it will be even easier to copy all resources from upstream.

vrozkovec:
Ok, in that case I'll also update to most recent 4.x version and do another PR

PR #834, opened 2020-01-23
Wicket 9.x bootstrap 4.x

solomax:


solomax:
Old IE dependencies are removed

PR #839, opened 2020-01-27
Remove multiple="multiple" attribute if only one file can be uploaded.

vrozkovec:


martin-g:
Thank you, @vrozkovec !

PR #841, opened 2020-02-06
Date pickers seems to work and improved to work with LocalDate/LocalD…

solomax:
…ateTime


solomax:
Would appreciate if you can review the code :)
It seems to work now
and also seems to be more customizable :)

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/AbstractDateTimePickerWithIcon.java
```diff
@@ -116,4 +118,9 @@ protected Component newIcon(String wicketId) {
     protected IconType newIconType() {
         return FontAwesomeIconType.calendar;
     }
+
+    @Override
+    protected IMarkupSourcingStrategy newMarkupSourcingStrategy() {
```
martin-g: Maybe you can extend from `FormComponentPanel` instead ?!

bootstrap-samples/pom.xml
```diff
@@ -51,6 +51,12 @@
             <groupId>de.agilecoders.wicket</groupId>
             <artifactId>wicket-bootstrap-themes</artifactId>
         </dependency>
+        <!-- icons -->
+        <dependency>
```
martin-g: This dependency comes as transitive one from here: https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/c95661ed224909f28342e45ec11f1c71c5b19241/bootstrap-extensions/pom.xml#L61-L65
Just update it from 5.7.1 to 5.9.0 there

martin-g (changes requested):

solomax:
Done :)

martin-g:
Thank you, @solomax !

martin-g:
I've ported the improvements to `wicket-9.x`. I haven't tested them. I hope everything still works fine.

PR #843, opened 2020-02-07
Sub-components are added in onInitialize

solomax:


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/BootstrapButton.java
```diff
@@ -49,11 +49,18 @@ public BootstrapButton(final String componentId, final Buttons.Type type) {
     public BootstrapButton(final String componentId, final IModel<String> model, final Buttons.Type type) {
         super(componentId, model);
 
-        add(buttonBehavior = new ButtonBehavior(type, Buttons.Size.Medium));
+        buttonBehavior = new ButtonBehavior(type, Buttons.Size.Medium);
+
+        icon = newIcon("icon");
```
martin-g, line 54: This needs to be improved. Calling overrideable methods in the constructor is not good.
Do you want to improve it or should I do it ?

solomax: Would appreciate if you will do it :)

solomax: @martin-g there seems to be lots of places for improvement
Maybe you can you can show me what do you have in mind and I'll prepare PR?

martin-g: Sorry! I was busy with other things!
Improved it!

PR #844, opened 2020-02-07
Active for ajax tabbed is fixed

solomax:


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/tabs/AjaxBootstrapTabbedPanel.java
```diff
@@ -44,8 +45,9 @@ public void onClick(final Optional<AjaxRequestTarget> targetOptional) {
 				targetOptional.ifPresent(target -> target.add(AjaxBootstrapTabbedPanel.this));
 				onAjaxUpdate(targetOptional);
 			}
-
 		};
+		link.add(Behavior.onAttribute("class", old -> old + (index == getSelectedTab() ? " " + getSelectedTabCssClass() : "")));
```
martin-g: I think I've fixed this with https://github.com/l0rdn1kk0n/wicket-bootstrap/commit/c95661ed224909f28342e45ec11f1c71c5b19241
Is it wrong/broken ?

solomax: AjaxBootstrapTabbedPanel overrides method from BootstrapTabbedPanel without call to super

I can try to change my code to reuse your solution
(not sure temporary behavior will work)

solomax: @martin-g ping :)

martin-g:
Thank you, @solomax !

PR #846, opened 2020-02-10
Component adding is moved to onInitialize()

solomax:


martin-g:
Awesome!
Thank you, @solomax !

PR #847, opened 2020-02-11
Component adding is moved to onInitialize()

solomax:
@martin-g not sure if it is good idea to add ButtonBehavior in constructor but without it Navbar is broken :(
this code doesn't do what expected: https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/wicket-9.x-bootstrap-4.x/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/navbar/NavbarDropDownButton.java#L80

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Alert.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Alert.java
@@ -1,8 +1,6 @@
 package de.agilecoders.wicket.core.markup.html.bootstrap.dialog;
 
-import de.agilecoders.wicket.core.markup.html.bootstrap.behavior.BootstrapResourcesBehavior;
-import de.agilecoders.wicket.core.markup.html.bootstrap.behavior.ICssClassNameProvider;
-import de.agilecoders.wicket.core.util.Attributes;
+import java.time.Duration;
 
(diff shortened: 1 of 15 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Alert.java
```diff
@@ -85,21 +87,50 @@ public Alert(String id, IModel<String> message, IModel<String> header) {
...
-        };
+    private Label getBlockHeader() {
+        if (blockHeader == null) {
+            blockHeader = new Label("block", header);
+        }
+        return blockHeader;
+    }
 
-        add(this.inlineHeader, this.blockHeader, this.message, this.closeButton);
+    protected Component getMessage() {
```
martin-g: Does this need to be `protected` ?
The idea is to not call overridable methods in the constructor because their bodies may assume that the current instance is fully initialized.

solomax: Otherwise NotificationAlert get broken .... :(

martin-g: OK. Please add `final` to it then

martin-g:
Thank you!

PR #849, opened 2020-02-14
Label and model are set to internal date/time field

solomax:


martin-g:
Thank you, @solomax !

Issue #850, opened 2020-02-23
BootstrapFileInputField does not allow setting the icon for cancelling


reckart:
BootstrapFileInputField uses glyphicon by default, but that is not available by default with Bootstrap 4. While overriding the icons for most buttons, the icon for the cancel button cannot be set.

martin-g:
Do you like to send a Pull Request ?
Some third party widgets (like Bootstrap FileInput) may not have a version for Bootstrap 4 or may just need an update.
Help is welcome!

reckart:
I have the changes locally already - so yes, expect a PR for this one. Whatever I'm using atm in terms of widgets looks okish, although I had to set up a shim SCSS mapping some old Bootstrap 3 names to corresponding Bootstrap 4 classes.

reckart:
Change has been merged.

PR #851, opened 2020-02-27
#850 - BootstrapFileInputField does not allow setting the icon for cancelling

reckart:
- Add option for setting the cancel icon
- Adjusted icons set by Wicket code to correspond to default icons from the JS (at least for "remove" and "cancel")


martin-g:
Thank you, @reckart !

Issue #852, opened 2020-02-28
SassResourceReference does not work if SCSS files are in JARs


reckart:
It appears that JSASS only knows how to compile SCSS files that are residing on the file system, but not if they are located in JARs.

```
Caused by: io.bit3.jsass.CompilationException: Internal Error: File to read not found or unreadable: jar:file:/home/user/.m2/repository/de/tudarmstadt/ukp/clarin/webanno/webanno-ui-core/4.0.0-SNAPSHOT/webanno-ui-core-4.0.0-SNAPSHOT.jar!/de/tudarmstadt/ukp/clarin/webanno/ui/core/css/theme/bootstrap.scss

	at io.bit3.jsass.adapter.NativeAdapter.compileFile(Native Method) ~[jsass-5.10.3.jar:5.10.3]
	at io.bit3.jsass.adapter.NativeAdapter.compile(NativeAdapter.java:44) ~[jsass-5.10.3.jar:5.10.3]
	at io.bit3.jsass.Compiler.compile(Compiler.java:127) ~[jsass-5.10.3.jar:5.10.3]
	at io.bit3.jsass.Compiler.compileFile(Compiler.java:79) ~[jsass-5.10.3.jar:5.10.3]
	at de.agilecoders.wicket.sass.SassCacheManager.getCss(SassCacheManager.java:133) ~[wicket-bootstrap-sass-3.0.0-M13.jar:?]
	... 92 more
```

martin-g:
Fixed by merging your PR!

PR #853, opened 2020-02-28
#852 - SassResourceReference does not work if SCSS files are in JARs

reckart:
If the root SCSS resource is not on the file system, then load its content from its URL and compile it as a String. This is useful for compiling SCSS resources localed in JAR files. Local imports are not supported in such SCSS files, but they can usually be replaced with package imports.


chrisco484 (commented):
Is scssConcent a typo? Did you mean scssContent?

reckart:
indeed

martin-g:
Thank you, @reckart !

PR #856, opened 2020-03-11
fix active state of tabs

webD97:
In Bootstrap 4, active tabs need the .active class on the \<a>-Tag
instead of the \<li>-Tag (This seems to be Wicket's default behavior).
This commit removes the .active class from the \<li>-Tag and adds it
to the \<a>-tag in order to fix the styling of active tabs.

For more information, see https://getbootstrap.com/docs/4.0/components/navs/#tabs


martin-g:
Thank you, @webD97 !

webD97:
Cool, thanks for the quick merge! Is there already a plan when the next release will be?

martin-g:
3.0.0-M14 is on its way to Maven repo! Enjoy!

PR #859, opened 2020-04-24
BootstrapTabbedPanel has missing css attributes

igorky:
When you select a tab the 'active' class should appear on a link 'a' tag, but now I see it on a parent 'li' tag. That actually causes that no tabs are actually selected for a user.


martin-g:
Thank you for the PR, @igorky !
I've downported the implementation from wicket-9.x-bootstrap-4.x to keep them in sync.

Issue #862, opened 2020-06-16
Buttons.Type.Menu should not add .nav-link CSS class


webD97:
https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/wicket-8.x-bootstrap-4.x/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/Buttons.java#L92

Since Bootstrap 4 (https://github.com/l0rdn1kk0n/wicket-bootstrap/commit/7ef866513d96fa292470aac650f0a25f419849d3#diff-75f58d6c697e3697ac207d183d77a241), buttons created with a type of Menu receive a `.nav-link` CSS class. This class is not related to buttons at all (it belongs to [navs](https://getbootstrap.com/docs/4.0/components/navs/#base-nav)). It effectively makes the button `display: block` causing all sorts of layout issues.

Since I don't see why it was added there in the first place (buttons are something atomic and should not assume being used in a nav), I'd like to suggest to remove the class again. What are your thoughts on this?

martin-g:
Sounds similar to https://github.com/l0rdn1kk0n/wicket-bootstrap/pull/855
Please send a Pull Request!

webD97:
Yeah, I can do that. But I don't want to remove stuff without finding out why it was put there in the first place.

webD97:
I think we can close this now. Sorry for bothering you again with this question, @martin-g , but are there any plans for the next release date on the 3.x series?

martin-g:
3.0.0-M16 is on its way to Maven Central!

PR #863, opened 2020-06-24
Remove .nav-link CSS class from Buttons.Type.Menu

webD97:
As discussed in #862:

The .nav-link CSS class is meant to be used for navigation menus, see https://getbootstrap.com/docs/4.0/components/navs/. It should not be used on arbitrary buttons because it adds a `display: block` style.


martin-g (approved):

martin-g:
Thank you, @webD97 !

webD97:
Great, thank you! Are there any plans for the next release date?

Issue #866, opened 2020-08-17
Menu is broken in 5.0.0


solomax:
All menu links doesn't have `nav-link` CSS class, so the styling is broken
The issue is caused by this 1bda15b7411186cff1b08eb568aa8b539684a0f6 commit

webD97:
I removed the `.nav-link` class from Buttons.Type because it is not a "type of button" (https://getbootstrap.com/docs/4.5/components/buttons/#examples) but a class specific to building menus, see https://getbootstrap.com/docs/4.5/components/navs/#base-nav. I think we should not add the class in Buttons.Type again as it issemantically incorrect.

I did a quick search in the code base on the master branch and found the [MenuButton](https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/wicket-9.x/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/button/dropdown/MenuButton.java) class. Maybe it would be appropriate to add the `.nav-link` class there and then use this component in your code?

Edit: Hm, this class seems to be dropdown related so this might not be ideal as well.

solomax:
From what I can see `NavbarAjaxButton`, `NavbarAjaxLink`, `NavbarButton`, `NavbarExternalLink` are all sort of buttons of type `Menu`

I see no bit deal to have this special type of navigation button

Some bigger refactoring is required to make things better alligned with bootstrap ideology

solomax:
This issue blocks release for us
Can we agree for a quick fix, or should i revert to 5.0.0-M2?

webD97:
I think we should work on a proper solution here i.e. a solution that works for everyone. Would shadowing the class and adding the CSS class there be an option for your release?

solomax:
i'm with reverting to M2 :)

btw what problem are you solving with 1bda15b commit?

webD97:
Okay :) I submitted this change because the `.nav-link` CSS class was not present in the Bootstrap 3 variant of wicket-bootstrap. When we upgraded to a wicket-bootstrap release with Bootstrap 4, this class broke our layout (it basically adds `display: block` to the button if I remember correctly). We did not find out why it was added in the first place and assumed it being a bug.

solomax:
i see :)
OK, let's refactoring begins :)

solomax:
Would it be OK for you 
if we will add `Buttons.type.NavLink` and will use where it is appropriate?

mweirauch:
[...]

PR #867, opened 2020-08-19
Fixes issue #866: Navigations and drop-downs seems to work

solomax:


mweirauch:
Just as a quick headsup - don't have much time during the day. This PR breaks button groups. As you've fixed the CSP protection in the samples, you should be able to see it in the ComponentsPage. All buttons are assigned `dropdown-item` rendering them wrong. I applied the `dropdown-item` elsewhere, but don't have to time to explain in detail now.

Not directly related, but vertical button groups are also broken (not due to this patch). Likely due to the Boostrap4 migration. Apparently the must not be assigned both `btn-group-vertical` and `btn-group` together anymore, but just `btn-group-vertical`. (You can see the visual glitch at the top button corners and the lower buttons being left-shifted.)

Otherwise looks good so far from a quick glance!

solomax:
Closing this PR in favor of #868

PR #868, opened 2020-08-20
w9b4: Fix missing "nav-link" for Navbar links

### LINKED ISSUE #866: Menu is broken in 5.0.0

mweirauch:
This PR is based upon @solomax work in #867 with button-groups fixed.

The CSP fix for the samples application is a separate commit, so that the main fix could potentially be backported to w8b4.

Things look like this now:

image

Fixes #866

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/navbar/NavbarDropDownButton.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/navbar/NavbarDropDownButton.java
@@ -1,15 +1,10 @@
 package de.agilecoders.wicket.core.markup.html.bootstrap.navbar;
 
-import java.util.List;
-
 import org.apache.wicket.markup.html.WebMarkupContainer;
-import org.apache.wicket.markup.html.link.AbstractLink;
 import org.apache.wicket.model.IModel;
 
 import de.agilecoders.wicket.core.markup.html.bootstrap.behavior.CssClassNameAppender;
-import de.agilecoders.wicket.core.markup.html.bootstrap.behavior.CssClassNameRemover;
 import de.agilecoders.wicket.core.markup.html.bootstrap.button.ButtonBehavior;
-import de.agilecoders.wicket.core.markup.html.bootstrap.button.ButtonList;
 import de.agilecoders.wicket.core.markup.html.bootstrap.button.Buttons;
 import de.agilecoders.wicket.core.markup.html.bootstrap.button.dropdown.DropDownButton;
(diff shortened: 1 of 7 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/navbar/NavbarDropDownButton.java
```diff
@@ -66,23 +61,7 @@ protected final void addButtonBehavior(final ButtonBehavior buttonBehavior) {
...
     protected WebMarkupContainer newButton(String markupId, IModel<String> labelModel, final IModel<IconType> iconTypeModel) {
         WebMarkupContainer button = super.newButton(markupId, labelModel, iconTypeModel);
-        button.add(new CssClassNameRemover("btn"));
-        button.add(new CssClassNameAppender(Buttons.Type.Menu.cssClassName()));
+        button.add(new CssClassNameAppender(Buttons.Type.NavLink.cssClassName()));
         return button;
     }
-
-    @Override
-    protected ButtonList newButtonList(String markupId) {
```
solomax, line 75: I see design issue here: every class which will override `newButtonList` will have `dropdown-item` CSS class dropped :(

I had in plans to fix it as well ...

mweirauch: Ah I see, that's why you moved this to `ButtonList::configureLink`. Nevertheless the design issue was there before and after realising this code is "useless" in the sense `Buttons.Type.Menu` is not providing anything meaningful anymore, I opted to drop it entirely.

Perhaps you can come up with a better design after the main issues are fixed?

solomax: sure, it is not blocker here
I wrote it to not forget :) (sort of "notes to myself" :)) )

solomax:
just was able to test everything locally (sorry for delay :( )
Everything seems to be great :)
@martin-g can you check and merge this PR?

martin-g (approved):

martin-g:
Thank you, @mweirauch & @solomax !
I've backported it to w-8.x-b-4.x

martin-g:
5.0.1 has been released!

solomax:
Thanks for the release @martin-g !

PR #869, opened 2020-08-22
w9b4: Fix CSS for vertical button groups

mweirauch:
Only up to Bootstrap 2.x "btn-group" and "btn-group-vertical"
where needed for vertical orientation. Since Bootstrap 3.x
only the latter is required.

https://getbootstrap.com/docs/4.4/components/button-group/#vertical-variation
https://getbootstrap.com/docs/3.4/components/#btn-groups-vertical
versus
https://getbootstrap.com/2.3.2/components.html#buttonGroups

Before: w9b4-vbg-before After: w9b4-vbg-after

Can likely be backported to w8b4 und w8b3. Haven't checked, though!


martin-g (approved):

martin-g:
Thank you, @mweirauch !
I've ported it to all Bootstrap 3.x and 4.x branches

PR #875, opened 2020-10-10
Fix FontAwesome5Graphic interface to have public access modifier

ursjoss:
Hi @martin-g . While converting parts of a project using wicket-bootstrap to kotlin, I ran into this issue with the default access modifier for the interface `FontAwesome5Graphic`. If it is not too much effort, I would highly appreciate a release with this fix, if it is not too much of an effort... Thanks upfront.


martin-g (approved):

martin-g:
Thanks, @ursjoss !
I will cut new releases soon! WicketStuff is being released at the moment.

PR #878, opened 2020-10-14
Fixed SummernoteEditor to retain user input on form validation error.

RLStokes:
SummernoteEditor loses the users input when another field in the form get a validation error and the submitted value has not been pushed to the model. The FormComponent getValue() method will return the submitted value. For values to not be escaped by getValue() SummernoteEditor setEscapeModelStrings(false) must be called after construction.


martin-g (approved):

martin-g:
Thank you, @RLStokes !

martin-g:
The new release is being uploaded to Maven Central!

PR #881, opened 2020-11-07
remove sizing of form groups as it is not supported by Bootstrap 4

pinhead84:
Hey guys,

Bootstrap 4 does not support sizing of Form Groups. The classes `form-group-sm` & `form-group-lg` are no more available in their CSS and its not available in their documentation anymore. Therefore this PR removes the corresponding code from the `FormGroup` class.

It might lead to compilation problems, if developers are still using `FormGroup.Size` in their application. Do you prefer using deprecation annotations instead of just removing this feature?

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroup.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroup.java
@@ -18,6 +18,8 @@ import org.apache.wicket.model.Model;
 import org.apache.wicket.util.string.Strings;
 import org.apache.wicket.util.visit.IVisit;
 import org.apache.wicket.util.visit.IVisitor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -52,6 +54,8 @@ public class FormGroup extends Border {
         }
 
     }
+
+    private static final Logger LOG = LoggerFactory.getLogger(FormGroup.class);
 
     private Component label;
     private Component help;
@@ -112,6 +116,7 @@ public class FormGroup extends Border {
      * {@link InputBehavior#size(InputBehavior.Size)} on the {@link FormComponent} instead.
      */
     public FormGroup size(final Size size) {
+        LOG.warn("Ignore form group resizing as it is not supported by Bootstrap 4.");
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroup.java
```diff
@@ -31,31 +30,11 @@
  */
 public class FormGroup extends Border {
 
-    /**
-     * Holder class for all possible form group sizes
-     */
-    public enum Size implements ICssClassNameProvider {
```
martin-g, line 37: Let's just deprecate the enum. It will be removed later, probably in a major version.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroup.java
```diff
@@ -101,17 +80,6 @@ public FormGroup useFormComponentLabel(boolean value) {
         return this;
     }
 
-    /**
-     * sets the size of form-group
-     *
-     * @param size the size to use
-     * @return this instance for chaining
-     */
-    public FormGroup size(final Size size) {
```
martin-g, line 110: deprecate

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroup.java
```diff
@@ -158,9 +126,6 @@ protected void onComponentTag(ComponentTag tag) {
 
         checkComponentTag(tag, "div");
         Attributes.addClass(tag, "form-group", stateClassName);
-        if (size != null) {
```
martin-g, line 161: Let's write a warning log here that size is ignored.

bootstrap-core/src/test/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroupTest.java
```diff
@@ -95,52 +95,6 @@ public void updateModel() {
         assertThat(updateModelCounter.get(), is(1));
     }
 
-    @Test
```
martin-g, line 98: This is OK to be removed!

martin-g (changes requested):
Hi @pinhead84 !

Thanks for the PR!
We cannot remove the public APIs in minor release. Let's do it with deprecation!

pinhead84:
Hey Martin,

thanks for the quick response and your notes. 

- I've restored the public API and marked it as deprecated with a plausible message.
- All private stuff regarding resizing is already removed.
- A warning is logged, if the public `FormGroup.size()` method is called.

Or do you prefer keeping the private API intact? - Maybe a developer still uses the `form-group-lg` / `form-group-sm` classes in his CSS definitions. Currently they are not written anymore, which may lead to rendering issues. In this case we might still keep the tests, even if the feature is deprecated and unsupported by Bootstrap 4.

Best regards.

martin-g (approved):

martin-g:
Thank you, @pinhead84 !

PR #882, opened 2020-11-08
fix validation on form groups for Bootstrap 4

pinhead84:
Bootstrap 4 does not support `is-valid` & `is-invalid` classes on form groups anymore. Instead `valid-feedback` & `invalid-feedback` classes are used on the feedback container (with `wicket:id="error"`). See <https://getbootstrap.com/docs/4.5/components/forms/#server-side>.

This PR 

- removes the `is-valid` & `is-invalid` classes from form groups.
- adds `valid-feedback` & `invalid-feedback` classes to the feedback container.
- removes the `form-text` class from the feedback container in `FormGroup.html` according to Bootstrap 4 documentation.
- replaces `<p>` elements for help text and feedback container with `<div>` according to Bootstrap 4 documentation.

There are no changes, that should break the existing Java API. Only in case a developer implements his own `FormGroup.FeedbackMessageToCssClassNameTransformer` class, it might lead to rendering issues as this class is now used for the feedback container instead of the form group itself.

### DIFF
--- a/bootstrap-core/src/test/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroupTest.java
+++ b/bootstrap-core/src/test/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroupTest.java
@@ -147,8 +147,6 @@ class FormGroupTest extends WicketApplicationTest {
         tester().getSession().setLocale(Locale.ENGLISH); // for the validation message
 
         Model<FormData> model = Model.of(new FormData());
-        model.getObject();
-
         Form<FormData> form = new Form<>("form", new CompoundPropertyModel<>(model));
 
         FormGroup group = new FormGroup("id");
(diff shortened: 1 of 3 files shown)


bootstrap-core/src/test/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroupTest.java
```diff
@@ -166,7 +166,39 @@ void formGroupSubmitValidation() {
...
         tester().assertLabel("form:id:error", "&#039;value&#039; is required.");
-        tester().assertContains("class=\".*is-invalid.*\""); //assert error CSS class is present
+        tester().assertContains("class=\"invalid-feedback\""); //assert error CSS class is present
+    }
+
+    @Test
+    void formGroupSubmitValidationSuccess() {
+
+        Model<FormData> model = Model.of(new FormData("test"));
+        model.getObject();
```
martin-g: Is this call really needed here ?

martin-g (approved):

bootstrap-core/src/test/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroupTest.java
pinhead84: I also wasn't sure about the model creation and `model.getObject();`call - but just copied the code from the other tests in `FormGroupTest.java` and added the required adjustments for success validation (`new FormData("test")`).

I can simplify the model creation, if you like. But adjusting the other tests, that don't belong to form group validation, might be better placed in a separate PR?

pinhead84:
Hey Martin,

I've updated the PR:

- Using `java.util.function.Function` instead of `com.google.common.base.Function`. This might lead to problems for developers in the rare case, that they are implementing their own `FeedbackMessageToCssClassNameTransformer` for a form group. I don't know, if this is a problem for you.
- The unnecessary call to `model.getObject();` was removed in both validation tests. But other form group tests still contain this call. I'm not sure, if I should also remove them as the doesn't really belong to this PR.

martin-g:
Looks perfect! Thank you, @pinhead84 !

The other tests could be improved later. The extra call to `model.getObject()` seems unnecessary but also harmless.

PR #883, opened 2020-11-08
add validation classes to form controls for Bootstrap 4

pinhead84:
Bootstrap 4 uses the classes `is-valid` & `is-invalid` on form controls for server side validation. See <https://getbootstrap.com/docs/4.5/components/forms/#server-side>

This PR updates the `InputBehavior` accordingly - similar to the mechanism used in `FormGroup`. 

- If the `FormComponent` contains a success or info feedback message, the `is-valid` class is appended.
- If the `FormComponent` contains a more severe feedback message, the `is-invalid` class is appended.

This fix works nicely together with my other PR #882.

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/InputBehavior.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/InputBehavior.java
@@ -1,11 +1,17 @@
 package de.agilecoders.wicket.core.markup.html.bootstrap.form;
 
+import com.google.common.base.Function;
 import de.agilecoders.wicket.core.markup.html.bootstrap.behavior.BootstrapBaseBehavior;
 import de.agilecoders.wicket.core.markup.html.bootstrap.behavior.ICssClassNameProvider;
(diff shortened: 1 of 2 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/InputBehavior.java
```diff
@@ -1,12 +1,18 @@
 package de.agilecoders.wicket.core.markup.html.bootstrap.form;
 
+import com.google.common.base.Function;
```
martin-g: Some people complained of the usage of Guava classes. Can we use java.util.function.Function instead ?

pinhead84: I'd also prefer to avoid the Guava classes. I wasn't sure, if there is a deeper reason behind using the library and just didn't want to produce too many deviations from the existing code. I'm going to update the PR soon.

pinhead84:
Hey Martin,

I've updated the PR - now using `java.util.function.Function` instead of `com.google.common.base.Function`.

martin-g:
Danke, @pinhead84 !

PR #885, opened 2020-11-10
remove guava dependency for Wicket 9.x

pinhead84:
Hey Martin,

in one of the previous PR's you've suggested to avoid using Guava library classes. As I'm also not a big fan of this dependency and as it is used only rarely throughout the project, I've removed it entirely and replaced its usage with code from JDK and commons-lang.

Of course all tests are passing after the modification. Let me know what you think about it and feel free to merge it or not. Guava is not a showstopper for me - but I always like getting rid of unnecessary dependencies.

BTW: All modifications are compatible with JDK 8 / Wicket 8. But some conflicts conflicts occur on merge due to some differences in the code (e.g. order of imports). I resolved these issues and send you a separate PR #886 for Wicket 8.x.

### DIFF
--- a/bootstrap-core/pom.xml
+++ b/bootstrap-core/pom.xml
@@ -72,11 +72,6 @@
             <artifactId>wicket-util</artifactId>
         </dependency>
 
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-        </dependency>
-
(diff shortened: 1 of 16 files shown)


martin-g (approved):

martin-g:
Huge thanks, @pinhead84 !

PR #887, opened 2020-11-10
use "for" attribute in BootstrapCheckbox label

pinhead84:
Hey Martin,

the label written by `BootstrapCheckbox` does not contain a `for` attribute to assign the respective checkbox. This PR adds the missing attribute.


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/BootstrapCheckbox.java
```diff
@@ -62,7 +64,14 @@ protected void onInitialize() {
         super.onInitialize();
 
         add(getWrapper()
-                .add(newLabel("label", labelModel))
+                .add(newLabel("label", labelModel)
+                    .add(new AttributeModifier("for", new Model<String>() {
```
martin-g, line 68: nit: you can use Java 8 lambdas to simplify: 

```
add(new AttributeModifier("for", () -> getCheckbox().getMarkupId(true)));
```

martin-g (approved):

PR #889, opened 2020-11-21
#888 - Illegal char in path on Windows

jcklie:
Fixes https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/888 . 
The issue seems to be that URIs pointing to jar contents are not valid paths.

I would prefer to use a URI Builder, but do not want to add a dependency just for that.

See also https://github.com/inception-project/inception/issues/1858

### DIFF
--- a/bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java
+++ b/bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java
@@ -185,6 +185,7 @@ class UrlImporter implements Importer {
 
     private Optional<Import> resolveLocalDependency(URI base, String url) {
         LOG.debug("Going to resolve an import from local dependency: {}", url);
+
         String importUrl = getAbsolutePath(base, url);
         Optional<Import> localImport = resolveLocalFileDependency(importUrl);
 
@@ -231,8 +232,8 @@ class UrlImporter implements Importer {
 
     private String getAbsolutePath(URI base, String url) {
         String basePath = base.toString();
-        Path parentBasePath = Paths.get(basePath).getParent();
-        return parentBasePath.resolve(url).toString();
+        String parentBasePath = basePath.substring(0, basePath.lastIndexOf('/'));
+        return parentBasePath + '/' + url;
     }
 
     private Import buildImport(URL importUri) {
--- a/bootstrap-sass/src/test/java/de/agilecoders/wicket/sass/BootstrapSassTest.java
+++ b/bootstrap-sass/src/test/java/de/agilecoders/wicket/sass/BootstrapSassTest.java
@@ -18,6 +18,7 @@ import org.junit.jupiter.api.Test;
 import java.net.URI;
 import java.net.URL;
 
+import static java.lang.System.lineSeparator;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
@@ -61,7 +62,7 @@ class BootstrapSassTest {
 
         SassSource sassSource = sass.getSassContext(res, null);
         String css = sass.getCss(sassSource);
-        assertThat(css, is(equalTo(".rule {\n  background: #999; }\n")));
+        assertThat(css, is(equalTo(".rule {" + lineSeparator() + "  background: #999; }" + lineSeparator())));
     }
 
     /**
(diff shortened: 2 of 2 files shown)


jcklie:
You might want to add Windows to your CI or switch to Github actions where that seems to be easier.

bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java
```diff
@@ -231,8 +232,8 @@ private String addUnderscore(String url) {
 
     private String getAbsolutePath(URI base, String url) {
         String basePath = base.toString();
-        Path parentBasePath = Paths.get(basePath).getParent();
-        return parentBasePath.resolve(url).toString();
+        String parentBasePath = basePath.substring(0, basePath.lastIndexOf('/'));
```
martin-g, line 235: I don't have Windows to test it but why `lastIndexOf('/')` ? Shouldn't it be `\` for Windows ?

bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java
```diff
@@ -231,8 +232,8 @@ private String addUnderscore(String url) {
...
+        return parentBasePath + '/' + url;
```
martin-g, line 236: Same

bootstrap-sass/src/test/java/de/agilecoders/wicket/sass/BootstrapSassTest.java
```diff
@@ -61,7 +62,7 @@ void importWebContext() throws Exception {
 
         SassSource sassSource = sass.getSassContext(res, null);
         String css = sass.getCss(sassSource);
-        assertThat(css, is(equalTo(".rule {\n  background: #999; }\n")));
+        assertThat(css, is(equalTo(".rule {" + lineSeparator() + "  background: #999; }" + lineSeparator())));
```
martin-g, line 65: Using `lineSeparator()` looks better than the hardcoded `/`s in UrlImporter

bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java
```diff
@@ -231,8 +232,8 @@ private String addUnderscore(String url) {
 
     private String getAbsolutePath(URI base, String url) {
         String basePath = base.toString();
-        Path parentBasePath = Paths.get(basePath).getParent();
```
martin-g, line 234: IMO this should be `Paths.get(base)`, i.e. use the `URI`, instead of `String`.
Could you please test it on Windows ?

bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java
jcklie: This is a URL not a path, so it is also a front slash on Windows.

bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java
jcklie: If you do that, then you get `Method threw 'java.nio.file.FileSystemNotFoundException' exception.`

bootstrap-sass/src/test/java/de/agilecoders/wicket/sass/BootstrapSassTest.java
jcklie: The hardcoded "/" is there because it is a URL, similar to https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/wicket-9.x-bootstrap-4.x/bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java#L209

jcklie:
If you want, then I add you Github Workflows also running on Windows, that can run parallel to your travis.

bootstrap-sass/src/main/java/de/agilecoders/wicket/sass/UrlImporter.java
martin-g: That would be awesome!

martin-g:
> If you want, then I add you Github Workflows also running on Windows, that can run parallel to your travis.

That would be awesome!

jcklie:
How quick can you release this btw? It is blocking me sadly from doing anything with INCEpTION on windows right now.

martin-g:
I've started the release!
It should be at Maven Central in an hour.

PR #895, opened 2021-02-25
Fix ConfirmationBehavior when used with Bootstrap 3.4.1

ursjoss:
This PR fixes the `ConfirmationBehavior` that without it does not display the Yes/No Buttons when used with bootstrap 3.4.1 (no problem with bs 3.4.0 though).

The bug was fixed upstream in this [ticket](https://github.com/mistic100/Bootstrap-Confirmation/issues/100).

This PR replaces `bootstrap-confirmation.js` (incl. the minified version) with the current version from upstream: 2.4.4.

The SNAPSHOT version of wicket-bootstrap built from this PR works fine in my project (see [ticket](https://github.com/ursjoss/scipamato/issues/262))


martin-g:
Thank you, @ursjoss !

ursjoss:
Thank you @martin-g 😊

PR #898, opened 2021-04-21
Fixing bootstrap and bootswatch CDN URLs, Updating themes to 4.6.0

mwhesse:


bootstrap-core/src/main/java/de/agilecoders/wicket/core/settings/BootstrapSettings.java
```diff
@@ -57,8 +57,8 @@ public BootstrapSettings() {
     }
 
     @Override
-    public IBootstrapSettings setVersion(String version) {
-        this.version = version;
+    public IBootstrapSettings setBootstrapVersion(String version) {
```
martin-g: We cannot break the API in minor releases. Please revert.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/settings/IBootstrapSettings.java
```diff
@@ -11,10 +11,15 @@
...
+    /**
+     * The version of Bootstrap for Webjars
+     */
+    String BOOTSTRAP_WEBJARS_VERSION = "4.6.0-1";
+    
     /**
      * The version of Bootstrap
      */
-    String VERSION = "4.6.0-1";
+    String BOOTSTRAP_VERSION = "4.6.0";
```
martin-g: Please revert the constant renaming.

martin-g: Why do we have now two versions - plain and webjars ? Can we use just `4.6.0` ?

martin-g (changes requested):

bootstrap-core/src/main/java/de/agilecoders/wicket/core/settings/IBootstrapSettings.java
mwhesse: There was a fix release "4.6.0-1" for the bootstrap 4.6.0 webjar. The  BOOTSTRAP_WEBJARS_VERSION constant is used in the BootstrapJsReferenceTest

bootstrap-core/src/main/java/de/agilecoders/wicket/core/settings/BootstrapSettings.java
mwhesse: Done

martin-g (approved):

martin-g:
Thank you, @mwhesse !

PR #900, opened 2021-05-19
Add Bootstrap Italia theme

geel97:
I implemented [bootstrap-italia theme](https://github.com/italia/bootstrap-italia/) inside Wicket. Still need a bit of testing. Unfortunatelly bootstrap-italia has fonts path written with absolute path instead of a relative path. For this reason the folder  [`fonts`](https://github.com/geel97/wicket-bootstrap/tree/wicket-9.x/bootstrap-themes/src/main/java/de/agilecoders/wicket/themes/markup/html/italia/fonts) must be moved in `/bootstrap-italia/dist`. I tried using Wicket's `ResourceReference` but no luck.
I will do some more tests.


bootstrap-themes/src/main/java/de/agilecoders/wicket/themes/markup/html/italia/ItaliaCssReference.java
```diff
@@ -0,0 +1,27 @@
+package de.agilecoders.wicket.themes.markup.html.italia;
+
+import org.apache.wicket.request.resource.CssResourceReference;
+
+public class ItaliaCssReference extends CssResourceReference {
+	
+	private static final long serialVersionUID = 4361205879317055824L;
+
+    public ItaliaCssReference() {
+        super(ItaliaCssReference.class, "css/bootstrap-italia.min.css");
```
martin-g: No need to have `.min` in the file name. Wicket will use the `.css` file in DEVELOPMENT mode and the `.min.css` in DEPLOYMENT mode

bootstrap-themes/src/main/java/de/agilecoders/wicket/themes/markup/html/italia/ItaliaTheme.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public ItaliaTheme() {
+        this("italia");
+    }
+    
+    @Override
+    public List<HeaderItem> getDependencies() {
+        List<HeaderItem> references = new ArrayList<HeaderItem>();
+        references.add(CssHeaderItem.forReference(ItaliaCssReference.instance()).setId(BOOTSTRAP_THEME_MARKUP_ID));
+        references.add(JavaScriptHeaderItem.forReference(new JQueryPluginResourceReference(ItaliaTheme.class, "js/bootstrap-italia.min.js")));
```
martin-g: No need of `.min`. See above

martin-g (commented):
1. Is it possible to add also the non-minified files ?
2. Check this project sources for `gettheme.sh` files. Those are used to fetch the latest versions of the themes. It would be good if there is such for Italia too. Maybe also you could use it to put fonts/styles/scripts in nice file structure. You could also process the, - e.g. replace absolute paths with relative ones.

geel97:
Thanks to the Webjars, the implementation of [bootstrap-italia theme](https://github.com/italia/bootstrap-italia/) is now much easier.

geel97:
Bootstrap-italia require bootrstrap 4.6.1, so I think this is the wrong branch...
I close this PR and open another on the correct branch.

PR #908, opened 2021-06-21
Migration to bootstrap5 (incomplete)

solomax:


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/block/prettyprint/PrettifyJavaScriptReference.java
```diff
@@ -34,7 +35,7 @@ private PrettifyJavaScriptReference() {
 
     @Override
     public List<HeaderItem> getDependencies() {
-        List<HeaderItem> dependencies = Generics2.newArrayList(super.getDependencies());
+        List<HeaderItem> dependencies = super.getDependencies().stream().collect(Collectors.toCollection(ArrayList::new));
```
martin-g: `new ArrayList(super.getDependencies())` ?!

solomax: thanks :) `Generics2` looks unnecessary to me, but I stuck in couple of places :(

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
```diff
@@ -1,6 +1,6 @@
 package de.agilecoders.wicket.extensions.markup.html.bootstrap.editor;
 
-import de.agilecoders.wicket.extensions.markup.html.bootstrap.icon.FontAwesome5CssReference;
+import de.agilecoders.wicket.extensions.markup.html.bootstrap.icon.FontAwesomeCssReference;
```
martin-g, line 3: even that we drop FA4 I think it would be better to keep `5` in the name
I think it would be easier for the users to upgrade from Wicket-Bootstrap-4.x to upgrade to W-B-5.x

martin-g (approved):

martin-g:
I will deploy 6.0.0-SNAPSHOT on every merge from now on!
https://oss.sonatype.org/content/repositories/snapshots/de/agilecoders/wicket/wicket-bootstrap-core/6.0.0-SNAPSHOT/

solomax:
Thanks @martin-g :)
Will try to find more time on this

BTW `Bootstrap Confirmation` seems to be incompatibe with Bootstrap5
Shall we 

1. find substitute
2. create polyfill for `$.fn.popover`
3. drop this plugin

martin-g:
The best would be to convince the project maintainer to update it for 5.x
The second best is to find a substitution.
I prefer not to maintain patches for third party plugins.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
solomax: will rename back in next PR :)

PR #910, opened 2021-07-03
More work on modal, input-group-prepend/input-group-append are eliminated

solomax:


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Modal.java
```diff
@@ -529,7 +514,10 @@ public void renderHead(IHeaderResponse response) {
      * @see #createInitializerScript
      */
     protected String createBasicInitializerScript(final String markupId) {
-        return "$('#" + markupId + "').modal({keyboard:" + useKeyboard() + ", show:" + showImmediately() + "})";
+        return "{" //scope
+                + "const myModal = new bootstrap.Modal(document.getElementById('" + markupId + "'));" // options are added as data-attributes
+                + (showImmediately() ? "myModal.show();" : "")
```
martin-g, line 519: There is no need to create `myModal` if `showImmediately()` returns `false`.

martin-g (approved):

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Modal.java
solomax: should be addressed in next PR :)

PR #916, opened 2021-08-16
Use JQuery's hide() method on the component so there is no gap in DOM

vrozkovec:
If you add class `invisible`, component disappears but still takes up space, breaking the form layout.


martin-g:
Thank you, @vrozkovec !

PR #919, opened 2021-09-01
Date picker and confirmation are fixed

solomax:


solomax:
@martin-g Would it be OK to drop `jquery-migrate`

martin-g:
> Would it be OK to drop `jquery-migrate`

Sure!

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePickerBehavior.java
```diff
@@ -59,8 +59,8 @@ public void renderHead(Component component, final IHeaderResponse response) {
     @Override
     public void onComponentTag(Component component, ComponentTag tag) {
         super.onComponentTag(component, tag);
-        Attributes.set(tag, "data-bs-target", "#" + component.getMarkupId());
-        Attributes.set(tag, "data-bs-toggle", "datetimepicker");
+        Attributes.set(tag, "data-target", "#" + component.getMarkupId());
```
martin-g, line 62: Is this correct ?
I thought BS 5.x now uses `data-bs-**` everywhere ?!
Or maybe the datetimepicker is not yet migrated to BS 5.x ? In that case let's set both `data-target` and `data-bs-target`. Hopefully this will make it future-proof.

martin-g:
I see `prettify.js` being removed but don't see `prettify.css`.
Actually how did we use prettify-js in wicket-bootstrap ? Why it is no more needed ? I need such info in the PR/issue description when something is removed/added.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePickerBehavior.java
solomax: Actually datepicker is still BS-4 based
so it expects `data-*` without `bs` :(((
found it at our main app ....

solomax:
these `prettify*` resources has comment:
```
@deprecated Use {@link de.agilecoders.wicket.core.markup.html.bootstrap.block.prettyprint.PrettifyCssResourceReference} instead
 */
```
so I just remove outdated file
will check where `prettify.css` might be :)

solomax:
Hopefully all comments are addressed :)

martin-g (approved):

martin-g:
Thanks, @solomax !

New -SNAPSHOT or M4 ?

solomax:
Let it be SNAPSHOT for now :)
Will do some more testing

martin-g:
Done!

PR #927, opened 2021-09-14
fix: [#925] Downgrade closure-compiler-unshaded from v20210601 to v20200830

### LINKED ISSUE #925: current closure-compiler-unshaded version has guava, gson, javax, ant, arg4j in the jar
the last version without them seems to be v20200830

ursjoss:
Fixes #925.

closure-compiler-unshaded v20200830 is the last version (verified until current 20210907) not
containing gson. Downgrading to the last good known version to avoid issues from PR #797.

Please also port to `wicket-9.x-bootstrap-4.x


martin-g (approved):

martin-g:
Thank you, @ursjoss !

Issue #924, opened 2021-09-14
Popovers do not work when BS5 JQuery integration is disabled


reckart:
Bootstrap 5 no longer needs JQuery. To avoid Bootstrap 5 features interfering with other JQuery plugins, one can disable the JQuery Bootstrap integration on a per-page level by adding the attribute `data-bs-no-jquery="data-bs-no-jquery"` to the page body.

However, doing so breaks at least the Wicket Bootstrap PopoverBehavior because it uses `$(component).chain("popover", config).get()` to initialize the popover.

As much as possible, Bootstrap functionalities should be utilised without resorting to JQuery.

Cf. https://getbootstrap.com/docs/5.1/components/popovers/
Cf. https://getbootstrap.com/docs/5.1/getting-started/javascript/#jquery-events

martin-g:
We know that BS 5 does not use jQuery!
We need PRs, not issues! :-)

reckart:
Here you go: https://github.com/l0rdn1kk0n/wicket-bootstrap/pull/929

PR #929, opened 2021-09-15
Popovers do not work when BS5 JQuery integration is disabled (#924)

reckart:
- Change script snipped from using JQuery to using native Bootstrap


martin-g (approved):

martin-g:
Thank you, @reckart !

PR #930, opened 2021-09-23
Bump BootstrapCheckBoxPicker to v2.0

vrozkovec:
- config contains all options from https://vsn4ik.github.io/bootstrap-checkbox/
- update breaks some old config methods, but these should be easy to fix
- v 2.0 is BS4 compatible


bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/pages/CheckboxesPage.java
```diff
@@ -51,10 +52,10 @@ private void addCheckboxPicker() {
 
         BootstrapCheckBoxPickerConfig config = new BootstrapCheckBoxPickerConfig();
         config
-            .withOnClass("btn-info").withOffClass("btn-warning")
-            .withOnIcon(FontAwesome5IconType.thumbs_up_s).withOffIcon(FontAwesome5IconType.thumbs_down_s)
+            .withOnCls(Type.Info).withOffCls(Type.Warning)
+            .withOnIconCls(FontAwesome5IconType.thumbs_up_s).withOffIconCls(FontAwesome5IconType.thumbs_down_s)
             .withReverse(true)
-            .withStyle(ButtonGroup.Size.Small);
+            .withGroupCls(ButtonGroup.Size.Small);
```
martin-g, line 58: Could we preserve the old `with**` methods as `Deprecated` for while ? 
If they can delegate to the new ones it would be awesome! Otherwise they could throw an exception explaining what replaces them.

vrozkovec: @martin-g sure

vrozkovec: @martin-g Done!

martin-g (approved):

martin-g:
Thank you, @vrozkovec !

PR #931, opened 2021-09-28
Check if component is enabled in hierarchy

vrozkovec:
Without this change, it could happen that although component is disabled because of some parent getting disabled, UI did not reflect that.


pom.xml
```diff
@@ -565,16 +565,19 @@
...
-            <id>sonatype-nexus-staging</id>
-            <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
-        </repository>
-    </distributionManagement>
+
+<distributionManagement>
+    <repository>
+        <id>deploy-repo</id>
+        <name>Locally deployed releases</name>
+        <url>https://artifactory.rozkovec.name/artifactory/libs-release-local</url>
```
martin-g: Are you sure ? :-)

vrozkovec: Oops. That wasn't meant to be :)
Fixed.

martin-g (approved):

martin-g:
Thank you, @vrozkovec !

PR #933, opened 2021-11-01
BootstrapGenericPanel - set default model to topImage to prevent lookup

vrozkovec:
Without the model, Wicket always traversed the hierarchy up to find a model it could use.


martin-g:
Thank you, @vrozkovec !

PR #949, opened 2022-05-29
Make FileInputConfig maxFileSize and minFileSize aware

ursjoss:
While it's technically possible to add e.g. the "maxFileSize" configuration via `.getConfig().put(...)` to the `FileInputConfig`, it's IMHO a nice addition to make the config aware of that parameter. Added for symmetry reasons also "minFileSize".

As a side change, all paramters now also have a getter and the locales path in a javadoc comment is fixed.

### DIFF
--- a/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/fileinput/FileInputConfig.java
+++ b/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/fileinput/FileInputConfig.java
@@ -39,6 +39,10 @@ public class FileInputConfig extends AbstractConfig {
 
     public static final IKey<Integer> MinFileCount = newKey("minFileCount", 0);
 
+    public static final IKey<Integer> MaxFileSize = newKey("maxFileSize", 0);
+
+    public static final IKey<Integer> MinFileSize = newKey("minFileSize", 0);
+
     public static final IKey<String> BrowseIcon = newKey("browseIcon", "<i class=\"glyphicon glyphicon-folder-open\"></i> &nbsp;");
 
     public static final IKey<String> RemoveIcon = newKey("removeIcon", "<i class=\"glyphicon glyphicon-trash\"></i> &nbsp;");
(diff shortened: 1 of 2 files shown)


bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/fileinput/FileInputConfig.java
```diff
@@ -137,6 +141,18 @@ public FileInputConfig maxFileCount(int maxFileCount) {
         return this;
     }
 
+    /** Maximum file size for upload in KB - 0 (default) for no restriction on maximum file size */
+    public FileInputConfig maxFileSize(int maxFileSize) {
+        put(MaxFileSize, maxFileSize);
```
martin-g: Do we need to care about negative values ?

chrisco484: > Do we need to care about negative values ?

Isn't an upload of a negative file size actually, mathematically speaking, just a download? Lol :)

martin-g: Would it be better to use wicket-util's `Bytes` instead of `int` as type of the parameter ?
This will also take care of the negative values.

ursjoss: I was considering some kind of validation regarding negative values but refrained from it due to the fact that we don't validate other parameters either (e.g. maxFileCount). I didn't know about the `Bytes` implementation - I'll have a look at that option...

ursjoss: I changed the paramter for max/min FileSize to `Bytes`.

martin-g (approved):

martin-g:
Thank you, @ursjoss !

PR #944, opened 2022-02-11
Dependencies are updated according to release

solomax:

PR #950, opened 2022-05-29
Add font-awesome-6 and make it configurable to switch to 5

ursjoss:
Completely migrating to FontAwesome-6 for `wicket-9.x-bootstrap-5.x`.

I initially wanted to fully bump all the way to 6.1.1, but that version is not on maven-central for some reason. So I downgraded to 6.1.0 there. The CSS resource references still point to 6.1.1 though. Not sure what's best to do here... nor why 6.1.1 is not on maven central.

Also pulls along the sample pages, adjusts the footer to not mention glyph-icons but rather font-awesome.

Also downgrades jetty-maven-plugin, as I couldn't run the sample pages without that. Didn't investigate the root cause though.


martin-g:
I am not sure whether the removal of 5.x would break someone's application in a minor release.
@solomax Your opinion ?

reckart:
The upgrade from FontAwesome 4 to 5 was quite a bit of work in my app. I'm using the abstractions provided by wicket-bootstrap only in some places and in other (most) places, the glyphs are hardcoded into the HTML templates. Being forced to upgrade might be a bit annoying... that said... eventually one will have to upgrade...

solomax:
`4 -> 5` upgrade was flawless to me
AFAIR we had an agreement to leave only one version
In such case `5` suffix will only complicate things

@martin-g I would drop suffix :)

reckart:
The FA glyph prefixes had changed from 4 to 5. Also some glyphs went away and had to be replaced with others. Finally, widgets such as the bootstrap file upload widget had to be tweaked to work with the new version. :(

solomax:
yep :(
tweaks are required :( and it's painful from developer point of view :(

ursjoss:
Replacing the FontAwesome 5 classes was what I took out of the discussion in PR #944 - of course only in the `wicket-9.x-bootstrap-5.x` branch, which is not finalized yet. I suggest to leave the Index 6 in the class names nevertheless. That may be helpful in the future, when keeping version 6 and in addition optionally provide version 7 or later of FontAwsome...

But let me know if and how to adjust the PR and I'll have a look.

martin-g:
Maybe we should leave both set of classes for each version.
Also the Maven dependencies should become `optional`.
This way the users should decide which version to use in their apps.
Would that work ?

solomax:
The code will be duplicated a lot
I thought it would be good idea to have latest versions of everything :)
Sounds like It's better to have icon pack as separate project :)
old ones (4 and 5 for now) will be rarely updated

vrozkovec:
If possible, I'd like to be able to pull in FA version of my choosing. For example I'll stick with FA 5 for some time now, because of PRO version that I've purchased with a one-time payment during their Kickstarter campaign. To have FA 6 PRO version you'd need to pay yearly subscription which I kind of don't want to.

ursjoss:
I restored the FontAwesome5 classes and declared the maven depenency optional. Not sure if that captures the stakes of all involved in the discussion.

bootstrap-extensions/pom.xml
```diff
@@ -21,7 +21,7 @@
         <spin-js.version>2.1.0</spin-js.version>
         <animate.css.version>3.7.2</animate.css.version>
         <summernote.version>0.8.20</summernote.version>
-        <font-awesome.version>5.15.4</font-awesome.version>
+        <font-awesome.version>6.1.0</font-awesome.version>
```
solomax: we should ping someone to release `6.1.1` :)

pom.xml
```diff
@@ -62,7 +62,7 @@
         <!-- Note: closure-compiler -unshaded v20200920 (until at least v20210907) contains gson -->
         <google-closure.version>v20200830</google-closure.version>
         <javax.javaee-web-api.version>7.0</javax.javaee-web-api.version>
-        <jetty.version>11.0.8</jetty.version>
+        <jetty.version>9.4.44.v20210927</jetty.version>
```
solomax: `9.4.46.v20220331`

solomax (approved):

bootstrap-extensions/pom.xml
ursjoss: yes, of course. ~https://github.com/FortAwesome/Font-Awesome/issues/19092~ https://github.com/webjars/font-awesome/issues/70. Thanks, @solomax

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/AbstractDateTimePickerWithIcon.java
```diff
@@ -130,7 +130,7 @@ protected Component newIcon(String wicketId) {
      * @return icon type
      */
     protected IconType newIconType() {
-        return FontAwesome5IconType.calendar_alt_r;
+        return FontAwesome6IconType.calendar_r;
```
ursjoss: Here, we'd have a problem if the user uses derivses from this class (or uses (or with `DatetimePickerIconConfig` below) but chose to consume the other variant via maven coordinates, right? (Regardless if we use the 5 or 6 variant, it could go wrong with two versions)

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
```diff
@@ -90,7 +90,7 @@ public void onComponentTagBody(MarkupStream markupStream, ComponentTag openTag)
     @Override
     public void renderHead(IHeaderResponse response) {
         response.render(CssHeaderItem.forReference(SummernoteEditorCssReference.instance()));
-        response.render(CssHeaderItem.forReference(FontAwesome5CssReference.instance()));
+        response.render(CssHeaderItem.forReference(FontAwesome6CssReference.instance()));
```
martin-g: I think we need `IBootstrapSettings#fontAwesomeCssReference` + getter/setter

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/AbstractDateTimePickerWithIcon.java
martin-g: I didn't understand the problem here.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePickerIconConfig.java
```diff
@@ -15,15 +15,15 @@ public class DatetimePickerIconConfig extends AbstractConfig {
...
-    private static final IKey<String> Close = newKey("close", FontAwesome5IconType.times_s.cssClassName());
+    private static final IKey<String> Up = newKey("up", FontAwesome6IconType.arrow_up_s.cssClassName());
+    private static final IKey<String> Down = newKey("down", FontAwesome6IconType.arrow_down_s.cssClassName());
+    private static final IKey<String> Date = newKey("date", FontAwesome6IconType.calendar_days_r.cssClassName());
+    private static final IKey<String> Time = newKey("time", FontAwesome6IconType.clock_r.cssClassName());
+    private static final IKey<String> Previous = newKey("previous", FontAwesome6IconType.arrow_left_s.cssClassName());
+    private static final IKey<String> Next = newKey("next", FontAwesome6IconType.arrow_right_s.cssClassName());
+    private static final IKey<String> Today = newKey("today", FontAwesome6IconType.calendar_check_r.cssClassName());
+    private static final IKey<String> Clear = newKey("clear", FontAwesome6IconType.eraser_s.cssClassName());
+    private static final IKey<String> Close = newKey("close", FontAwesome6IconType.xmark_s.cssClassName());
```
martin-g: If we want to support more than one version of FA then we need more complex type in `IBootstrapSettings` than the one I suggested earlier ...
And here those are all `static` fields which makes it harder to even make use of the settings class ...

ursjoss:
Ok, to summon the plan and align my understanding:

- We extend IBootstrapSettings to accept the FontAwesome{5,6}CdnCssResource (defaulting to 6?)
- CheckboxX, SummernoteEditor, DateTimePicker etc would use the settings to get the reference of choice
- for the problem with the static IKeys in DatetimePickerIconConfig: we could set the default values in the static initialization to something irrelevant and actively set each value in the constructor from the settings...

In any way, IMHO, we'd need to propagate the font-awesome related classe sfrom bootstrap-extensions to bootstrap-core, right?

martin-g:
> We extend IBootstrapSettings to accept the FontAwesome{5,6}CdnCssResource (defaulting to 6?)

I think we will need something like `FontAwesomeSettings` that configures the CSS reference, the icon type, ...

martin-g:
> In any way, IMHO, we'd need to propagate the font-awesome related classe sfrom bootstrap-extensions to bootstrap-core, right?

Good point! 
IMO we should not move the classes but use the pattern from: 
 https://github.com/apache/wicket/blob/414d3f733e1d07a7f37f88d262b4a828612ab7ec/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/WebSocketSettings.java#L95-L119 (No need of `Holder` class since there are no fields in it).
This way any usage would look like `FontAwesomeSettings.get(Application.get()).getCssReference()`

ursjoss:
I tried to implement the suggested approach.

As I suggest to use FontAwesome-6 as a default, I removed the optional configuration in maven. We should have a non-optional dependency that can optionally be overriden if the user wants to do that.

I'm not particularly proud nor happy with the map for particular icons used in components inside of `wicket-extensions`. I'm probably overlooking a much simpler solution to allow dynamically picking the icons of the configured version of font-awesome. I'm happy to adjust that if you give me a hint.

ursjoss:
It is quite a quirk, IMHO.

On second thought, I'd say that `FontAwesomeSettings` should only be used inside of `wicket-extensions` with the exception of overriding the version in the application init method.

Spoiled from using kotlin, I wanted to set FontAwesomeSettings#getIconType and FontAwesomeSettings#IconKey `internal`, only making it available inside the maven module, but obviously that's not that easy in Java.

Client code (such as `wicket-samples`) should probably pick a FontAwesome version (5 or 6) and directly use the respective icons or css references. But wicket-extensions classes that do use font awesome icons can thus reflect the user choice of the version too.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/icon/FontAwesome5CDNCSSReference.java
```diff
@@ -4,21 +4,21 @@
 import org.apache.wicket.request.resource.UrlResourceReference;
 
 /**
- * <strong>Important</strong>: To use font-awesome 5.x you need to declare the Maven/Gradle dependency
```
solomax, line 7: these classes are for 5.x
It seems javadoc shouldn't be changed

ursjoss:
I fixed a bug with solid icons where the enum name starts with `_` (symbols 0-9), added some documentation and tried to remove some inadvertent whitespace changes to at least remove a tiny little bit of noise in the PR.

I will leave the PR as it is until we get some consolidated decision if and how to proceed.

bootstrap-extensions/pom.xml
ursjoss: font-awesome-6.1.1 is available in maven central now.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/AbstractDateTimePickerWithIcon.java
ursjoss: Components like this one used to depend on GlyphIcons that were bundled with bootstrap and therefore didn't need an additional dependency. PR #746 in 2018 switched from GlyphIcons to font-awesome icons (at the time to font-awesome-4.7.0, as the only version of font-awesome within wicket-bootstrap. The dependency on org.webjars:font-awesome:4.7.0 was non-optional, so the switch from bundled to non-bundled icons for this and similar components worked flawelessly.

Your initial suggestion for this PR was to make the dependency optional. If a user of wicket-bootstrap would not explicitly depend on font-awesome-6, trying to use the date time picker implementations would not work, as either non font-awesome version would be available at all or only version 5, if that was the users's choice.

The current approach of this PR treats the font-awesome dependency as non-optional, defaulting to version 6. If the user chooses to override the version to keep using version 5, the component selects the appropriate version of the icon.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/editor/SummernoteEditor.java
ursjoss: I did so according to my understanding.

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/datetime/DatetimePickerIconConfig.java
ursjoss: I wouldn't say it's solved, but at least worked around.

ursjoss:
As we now have font-awsome-6.1.1 in maven central, I think I've addressed all issues. Please comment on how to proceed.

solomax:
IMO it's time to merge this PR :)

martin-g:
Thanks, @ursjoss !
I will work on it later today!

ursjoss:
Thanks to all involved!

PR #953, opened 2022-06-28
fixes #951: configurable BootstrapBundle is introduced

### LINKED ISSUE #951: Bootstrap-confirmation is broken because of Popper
Latest `M7` release contains broken `confirmation`
After some digging I find-out this is because we have `popper` as additional dependency

The modern way is to drop it and use `bootstrap.bundle` instead
(see for ex. here https://stackoverflow.com/a/72336371)

Shall I proceed with such change?
Or Maybe it would be better (for whatever reason) have this behavior configurable?

@martin-g, @ursjoss WDYT? :)

solomax:
I would better drop all `popper.js` related code and keep `bootstrap.bundle` only :)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/references/BootstrapBundleJavaScriptReference.java
```diff
@@ -0,0 +1,36 @@
+package de.agilecoders.wicket.core.markup.html.references;
+
+import de.agilecoders.wicket.webjars.request.resource.WebjarsJavaScriptResourceReference;
+
+/**
+ * represents the bootstrap js library
+ *
+ * @author miha
```
martin-g: I am pretty sure `miha` does not care about this anymore :-)

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/references/BootstrapJavaScriptReference.java
```diff
@@ -45,7 +44,6 @@ private BootstrapJavaScriptReference() {
     @Override
     public List<HeaderItem> getDependencies() {
         return Dependencies.combine(super.getDependencies(),
-                                    JavaScriptHeaderItem.forReference(Application.get().getJavaScriptLibrarySettings().getJQueryReference()),
```
martin-g, line 48: Why ?

bootstrap-core/src/main/java/de/agilecoders/wicket/core/settings/BootstrapSettings.java
```diff
@@ -53,6 +57,7 @@ public BootstrapSettings() {
         this.updateSecurityManager = true;
         this.autoAppendResources = true;
         this.useCdnResources = false;
+        this.useBundle = false;
```
martin-g: Maybe we should switch to `true` as default. Otherwise your reported [issue](https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/951) would be still valid, no ?

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/references/BootstrapJavaScriptReference.java
solomax: As I understand from here: https://getbootstrap.com/docs/5.2/getting-started/introduction/
Bootstrap doesn't depends on jQuery anymore ...

bootstrap-core/src/main/java/de/agilecoders/wicket/core/settings/BootstrapSettings.java
solomax: Well,
popper version need to be fixed, via (additional) commit

IMO it worth to switch to bundle-only mode, this would make things simpler :)

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/references/BootstrapJavaScriptReference.java
martin-g: Good point!

bootstrap-core/src/main/java/de/agilecoders/wicket/core/settings/BootstrapSettings.java
martin-g: I agree with bundle-only setup!

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/references/BootstrapBundleJavaScriptReference.java
solomax: Addressed :)

bootstrap-core/src/main/java/de/agilecoders/wicket/core/settings/BootstrapSettings.java
solomax: Done :)

`MaterialDesignTheme` and `ItaliaTheme` still requires Popper (seems to)
but they are for bootstrap4 only :(

Shall we keep or drop it?

martin-g: I'd suggest to just remove these lines:
* https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/7010be70febdbe34d8adad3629b501056f767a8b/bootstrap-themes/src/main/java/de/agilecoders/wicket/themes/markup/html/italia/ItaliaTheme.java#L54
* https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/7010be70febdbe34d8adad3629b501056f767a8b/bootstrap-themes/src/main/java/de/agilecoders/wicket/themes/markup/html/material_design/MaterialDesignTheme.java#L32

solomax: done
Also dropped useless dependency :)

martin-g: https://github.com/l0rdn1kk0n/wicket-bootstrap/blob/7010be70febdbe34d8adad3629b501056f767a8b/bootstrap-themes/src/main/java/de/agilecoders/wicket/themes/markup/html/material_design/MaterialDesignTheme.java#L32
Do we still need this ?

solomax: this one is dropped already
will lead to compilation error otherwise :)

martin-g: Right! I am looking at the target branch ... 🤦🏻

martin-g (approved):

solomax:
I'd love to get release/new-snapshot as soon this one be merged :)

martin-g:
6.0.0-M8-SNAPSHOT has been uploaded to Sonatype!

solomax:
Thanks a million! :)

Issue #959, opened 2022-09-09
Bootstrap 4 Carousel Images are not scaled responsive


subes:
Bootstrap 3 (wicket-bootstrap 4.0.4):
![image](https://user-images.githubusercontent.com/332937/189359818-8d3d5fc1-68b0-4090-bcd5-9090303f42b0.png)

Bootstrap 4 (wicket-bootstrap version 5.0.6):
![image](https://user-images.githubusercontent.com/332937/189359921-bf3bb9b9-3b49-4711-a377-fbbfbc718627.png)

Looking at the documentation: https://getbootstrap.com/docs/4.6/components/carousel/
The img tag misses this attribute: class="d-block w-100". In bootstrap 3 the same was applied implicitly via the img-responsive that existed before. With bootstrap 4 one has to manually define this.

martin-g:
Please send a Pull Request!

subes:
Done, I also upgraded the bootstrap 5 carousel so that it works again.

PR #962, opened 2022-09-23
#959 make bootstrap 5 carousel responsive

subes:


subes:
I also updated the component to bootstrap 5 according to the documentation: https://getbootstrap.com/docs/5.2/components/carousel/

There is no need for JavaScript initialization anymore, because the data-bs-* attributes take care of that.

martin-g (approved):

PR #990, opened 2023-03-15
Make bootstrap-select compatible with bootstrap 5

keesvandieren:
bootstrap-select didn't work in Bootstrap 5.

The maintainer has created a beta-version of the bootstrap-select plugin which works.


keesvandieren:
It can be verified by running example page.

A release would be great :-)

bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/select/SelectJSReference.java
```diff
@@ -26,6 +26,6 @@ public static SelectJSReference instance() {
      * Private constructor.
      */
     private SelectJSReference() {
-        super(SelectJSReference.class, "js/bootstrap-select.js");
+        super("bootstrap-select/current/dist/js/bootstrap-select.js");
```
martin-g: Isn't there a `.min.js` in the webjar ?
Asking because you used the `min` version in SelectCSSReference.java

keesvandieren: @martin-g  good catch! Committed a fix

martin-g (approved):

martin-g:
Thank you, @keesvandieren !
6.0.2 is on its way to Maven Central!

PR #986, opened 2023-02-26
Issue #914: wicket-bootstrap-sass not working on Apple Silicon

reckart:
- Switch to ua.kasta:jsass which works on Apple Silicon

PR #1004, opened 2023-10-15
module-info are added; structrure is updated to conform M2

solomax:


bootstrap-sass/pom.xml
```diff
@@ -14,14 +14,12 @@
...
     <properties>
-        <autoModuleName>wicket.bootstrap.saas</autoModuleName>
-
         <antlr.version>3.5.2</antlr.version>
     </properties>
 
     <dependencies>
         <dependency>
-            <groupId>ua.kasta</groupId>
+            <groupId>io.bit3</groupId>
```
martin-g: I think we moved from `bit3` to `kasta` because `bit3` was abandoned and it does not support Mac ARM64.
See https://github.com/l0rdn1kk0n/wicket-bootstrap/pull/986

reckart: Mainly because of ARM - I believe it continues being abandoned in any case. The native library that is being wrapped here is no longer developed as far as I remember...

solomax: I'll revert this change

BUT according to https://mvnrepository.com/artifact/ua.kasta/jsass `ua.kasta:jsass` seems to be dead as well :(((
It is "one-release-repo" :((((

reckart: There is little point in maintaining a java wrapper for a library which is dead. See notice on jsass repo:

> libsass has been [declared deprecated](https://sass-lang.com/blog/libsass-is-deprecated) and there will be no further development! This also ends the further development of jsass. It is not yet clear whether there will ever be an implementation based on the [Sass embedded protocol](https://github.com/sass/embedded-protocol). I recommend switching to [Webpack](https://webpack.js.org/) or similar technologies! This project will continue to be maintained for the time being and compatibility with newer Java versions will be ensured. But please do not expect any further development on my part.

reckart: fwiw I have switched to using sass with the maven-frontend-plugin some time back.

martin-g (approved):

martin-g:
@solomax Anything else you want to add here or it is ready to be merged ?

solomax:
> @solomax Anything else you want to add here or it is ready to be merged ?

Let's wait for the M2 release to drop `SNAPSHOT` dependency :)
Meanwhile I'll try to create PRs for `jquery.selectors` and `wicket.webjars` so they will have Automatic module anmes (or be converted to JMS)

bootstrap-sass/pom.xml
solomax: > I have switched to using sass with the maven-frontend-plugin

I did the same :)

martin-g:
@solomax Ready to go ?

martin-g:
I wonder whether we need to add `open`s too ? Like WICKET-7081

solomax:
> I wonder whether we need to add `open`s too ? Like WICKET-7081

I'll try to check later this week or early next week :)

solomax:
> @solomax Ready to go ?

yes IMO this one can be released (so people can test it :)))

bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/WicketApplication.java
```diff
@@ -93,7 +111,7 @@ public void init() {
 
         optimizeForWebPerformance();
 
-        new AnnotatedMountScanner().scanPackage("de.agilecoders.wicket.samples.pages").mount(this);
```
solomax, line 96: Actually this class wasn't dropped: https://github.com/wicketstuff/core/blob/master/annotation/src/main/java/org/wicketstuff/annotation/scan/AnnotatedMountScanner.java

my local build is green ... :(

I can run `bootstrap-samples` using `mvn jetty:run`

```
Apache Maven 3.9.1 (2e178502fcdbffc201671fb2537d0cb4b4cc58f8)
Maven home: /opt/maven
Java version: 17.0.8.1, vendor: Private Build, runtime: /usr/lib/jvm/java-17-openjdk-amd64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.15.0-86-generic", arch: "amd64", family: "unix"
```

martin-g:
On Thu, 19 Oct 2023 at 19:11, Maxim Solodovnik ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In
> bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/WicketApplication.java
> <https://github.com/l0rdn1kk0n/wicket-bootstrap/pull/1004#discussion_r1365803784>
> :
>
> > @@ -93,7 +111,7 @@ public void init() {
>
>          optimizeForWebPerformance();
>
> -        new AnnotatedMountScanner().scanPackage("de.agilecoders.wicket.samples.pages").mount(this);
>
> Actually this class wasn't dropped:
> https://github.com/wicketstuff/core/blob/master/annotation/src/main/java/org/wicketstuff/annotation/scan/AnnotatedMountScanner.java
>
> my local build is green ... :(
>

Why green build is a bad thing ? :-)

I just removed the usage of AnnotatedMountScanner in WB.
Now there is no dependency to wicketstuff.

I can run bootstrap-samples using mvn jetty:run

solomax:
> Why green build is a bad thing ? :-) I just removed the usage of AnnotatedMountScanner in WB. Now there is no dependency to wicketstuff. I can run bootstrap-samples using mvn jetty:run

green build is not so bad :)
different results - is something need to be investigated IMO :))

martin-g:
Ah, you meant the issue with `maven-source-plugin`.
`mvn help:effective-pom` showed that there are two executions for this plugin. One of them comes from the parent pom (Sonatype OSS) - with id `attach-sources`

martin-g:
> yes IMO this one can be released

Done!

PR #1009, opened 2023-11-29
Use version defined in properties for maven-source-plugin

vrozkovec:


martin-g:
Thank you, @vrozkovec !

PR #1010, opened 2023-12-03
Replace .sr-only with .visually-hidden, use static Helper.screenReaderOnly class where applicable, fix SpinnerAjax(Button|Link) label rendering

vrozkovec:
In current implementation, just calling `setLabel()` triggers `AttributeModifier` that adds `.sr-only` even though the label's model may not be empty and in some setups label is not rendered because `.sr-only` class hides it. 

With this change, the check is made in `onConfigure`, but not very nice `get("label")` call is used - better way would require `BootstrapAjaxLink` and `BootstrapAjaxButton` methods that return label to be `protected` instead of `private`. Or extracting label's id to the constant and making protected that.

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/components/progress/Stack.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/components/progress/Stack.java
@@ -1,5 +1,6 @@
 package de.agilecoders.wicket.core.markup.html.bootstrap.components.progress;
 
+import org.apache.wicket.AttributeModifier;
 import org.apache.wicket.markup.ComponentTag;
 import org.apache.wicket.markup.head.IHeaderResponse;
 import org.apache.wicket.markup.head.OnDomReadyHeaderItem;
@@ -9,6 +10,7 @@ import org.apache.wicket.model.IModel;
 
 import de.agilecoders.wicket.core.markup.html.bootstrap.utilities.BackgroundColorBehavior;
 import de.agilecoders.wicket.core.util.Attributes;
+import de.agilecoders.wicket.core.util.CssClassNames.Helper;
 
 /**
  * Represents a stack of the progress bar.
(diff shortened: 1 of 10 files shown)


reckart:
If I remember correctly, in BS5, `sr-only` has been replaced: https://getbootstrap.com/docs/5.3/getting-started/accessibility/#visually-hidden-content

vrozkovec:
Right, I've missed that. I've slightly enlarged scope of this PR, replacing `.sr-only` with `.visually-hidden` and using static Helper, where applicable.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/components/progress/Stack.java
```diff
@@ -57,6 +59,7 @@ protected void onInitialize() {
         super.onInitialize();
 
         srOnly = new Label("srOnly", createLabelModel());
```
martin-g: Let's rename `srOnly` to `visuallyHidden` - to keep it consistent with the Bootstrap terminology

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/components/progress/Stack.java
```diff
@@ -57,6 +59,7 @@ protected void onInitialize() {
...
+        srOnly.add(AttributeModifier.append("class", Helper.screenReaderOnly));
```
martin-g: Same for `screenReaderOnly` - let's use the new terminology

vrozkovec:
All done

martin-g (approved):

martin-g:
Thank you, @vrozkovec !

martin-g:
Should we backport it to wicket-9.x-bootstrap-5.x ?

solomax:
> Should we backport it to wicket-9.x-bootstrap-5.x ?

yes, please :)

martin-g:
Done!

PR #1017, opened 2023-12-10
Wicket 10.x bootstrap 5.x FormGroup validation

vrozkovec:
See description of the problem in #1018


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/form/FormGroup.java
```diff
@@ -239,16 +240,50 @@ protected void onConfigure() {
         stateClassName = "";
         feedback.setDefaultModelObject("");
 
+        String validComponentClass = "is-valid";
+        String invalidComponentClass = "is-invalid";
```
martin-g, line 244: Those could be class constants

martin-g (approved):

martin-g:
Thank you, @vrozkovec !

PR #1021, opened 2023-12-10
Remove more usages of guava

ursjoss:
In #885, the dependency on guava was removed. This PR removes  some more leftover usages of guava.


martin-g (approved):

martin-g:
Thank you, @ursjoss !

Issue #1024, opened 2023-12-13
Bootstrap samples validation won't work


keesvandieren:
Reproduction:

* clone project, checkout wicket-10.x-bootstrap-5.x
* Import in IDEA, run `de.agilecoders.wicket.Start` in dir `$MODULE_WORKING_DIR$`
* Go to `http://localhost:8080/validation/tooltip`
* Submit the form

Exception stacktrace shows up:

```
	... 61 common frames omitted
2023-12-13 12:30:54,461 [qtp2003891312-21] WARN [org.eclipse.jetty.server.HttpChannel] - /validation/tooltip
java.lang.NoClassDefFoundError: com/google/common/base/Strings
	at de.agilecoders.wicket.extensions.markup.html.bootstrap.form.validation.FeedbackMessageBehavior.onComponentTag(FeedbackMessageBehavior.java:43)
	at org.apache.wicket.Component.renderComponentTag(Component.java:3961)
	at org.apache.wicket.Component.internalRenderComponent(Component.java:2501)
```

keesvandieren:
Guava is not on the classpath. It is a transitive dependency in bootstrap-extensions:

```
        <dependency>
            <groupId>com.google.javascript</groupId>
            <artifactId>closure-compiler-unshaded</artifactId>
            <optional>true</optional>
        </dependency>
```

Proposed solution: add guava as a direct dependency of bootstrap-extensions.

martin-g:
Fixed with e2cb5709

PR #1025, opened 2023-12-13
Add guava as dependency

keesvandieren:
Guava is missing but needed. It got unnoticed because it was delivered as transitive dependency of optional dependency in bootstrap-extensions

See #1024


martin-g:
No!
Guava should not be used at all! It brings only problems!

keesvandieren:
Thanks for fixing! de.agilecoders.wicket.extensions.markup.html.bootstrap.form.typeaheadV10.TypeaheadConfig also has reference to com.google.common.collect.ObjectArrays. Can you replace that too?

martin-g:
Done!

Issue #711, opened 2017-11-22
CodeBehavior does not show line numbers


martinmuellerrohde:
Trying to display line numbers using Code.setShowLineNumbers(true) doesn't work.
The rendered HTML shows a single L0 item instead of L0,L1,L2,... for every line of code.

This is causes by CodeBehavior.createInitializerScript() which changes the prettified HTML in line 119 by replacing linebreaks with "<br>":

`return "$('#" + markupId + "').html(PR.prettyPrintOne($('#" + markupId + "').html().replace(/\\r\\n|\\r|\\n/g,'<br>'), '', $('#" + markupId + "').attr('class')));";`

This bug can be fixed as follows:
`return "$('#" + markupId + "').html(PR.prettyPrintOne($('#" + markupId + "').html(), '', $('#" + markupId + "').attr('class')));";`

martin-g:
Thanks for the report!

Do you mind sending a Pull Request with the fix ?

PR #1027, opened 2024-01-12
Added missing caret for BootstrapSelectFilter. Improved ColorBehaviors

namtar:
# BoostrapSelectFilter
When using the BootstrapSelectFilter for example with a data table no caret is displayed for the DropDownChoice. This is because of a wrong css class set in the de/agilecoders/wicket/extensions/markup/html/bootstrap/table/filter/BootstrapSelectFilter.html markup file. Now a caret should be displayed.

# ColorBehaviors
I did some colorBehavior changes and added the ability to set opacity to texts or backgrounds and gradients to backgrounds, too. I changed existing classes and added new ones. Also I renamed Color enums. For example ColorBehavior.Color is now TextColorBehavior.TextColor.
There is an enum BackgroundColor for a BackgroundColorBehavior and a TextAndBackgroundColor for a TextAndBackgroundColorBehavior. The latter one is a behavior for https://getbootstrap.com/docs/5.3/helpers/color-background/ which determines automatically a contrasting textColor for a given background color.

The color changes may lead to compile errors on projects that use wicket-bootstrap and require mostly adjustments of class imports or the change of a method name. If this is a problem then reject this pull request and I will do a new one only for the BootstrapSelectFilter and CodeBehavior.

# CodeBehavior
I changed the initializerScript line in the CodeBehavior as proposed here (https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/711). Line numbers are shown again now, but since the google code-prettify project is archived perhaps it would be an idea to integrate the highlight.js library.


martin-g:
The PR is too big!
Maybe you could split it into several small ones?
Please try to not make formatting changes too!

vrozkovec:
It will also break all existing applications that use behaviors that were renamed, even though I like some names better, what's your approach in this case @martin-g?
I'd also perhaps keep behavior names so they match what Bootstrap has on their page `ColorBehavior` represents Utilities > Colors (renamed to `TextColorBehavior`), `ColorAndBackgroundBehavior` is for Helpers > Color & background (renamed to `TextAndBackgroundColorBehavior `), but as there are links in the header of the behavior that link to corresponding section, is it desirable to have names aligned like that..?

namtar:
Ok. No problem. I will split the pull request.

Is there a rule how the code formatting has to be done? I assumed that the .ediorconfig file is followed by Intellij IDEA and has the formatting rules that should be applied.

@vrozkovec I saw that Bootstrap has Utilities and Helpers sections and because of that I kept the matching behaviors in the corresponding packages. The word colors is very general though in this section (Utilities > Colors) they only add colors to text. Therefore I thought it would be a good idea to name the Enums and Behaviors as I have done in order to be more specific.
But because I have to split the PR anyway I can use the previous behavior names and only add gradients and opacity to the corresponding behaviors so that no existing application will break.

PR #1057, opened 2024-03-13
deps:  bump font awesome to 6.5.1

ursjoss:
dependabot updated the dependency in the pom file already. I created a separate PR #1059 to prevent it from doing that. Updating the version only will almost always be insufficient. We have to pull along the icon declarations as well.

This PR updates those to be consistent with version 6.5.1.


ursjoss:
@martin-g what would be the next step to get this merged?

martin-g:
Somehow I missed it ...
Thanks for the ping!

PR #1058, opened 2024-03-17
fix: Update regular/solid icons and brands to 5.15.4

ursjoss:
Adds some font awesome logos available in 5.15.4 and restores the trip-advisor icon I believed was removed in that version.


martin-g:
Thank you, @ursjoss !

PR #1073, opened 2024-04-10
Update BootstrapRadioGroup to support Bootstrap5 radio layout

x-paphio:
Bootstrap 5 radios are now with a Label having a "for" id attribute. This is different from the previous layout where the input was inside the label. (https://getbootstrap.com/docs/5.3/forms/checks-radios/#radios)

This was tested with bootswatch theme to fix the problem.


x-paphio:
need to fix an import

PR #1099, opened 2024-06-25
Update BootstrapRadioGroup.java

x-paphio:
My previous contribution somehow got the line adding the actual radio item to the group missing. Here adding it back and tested.


martin-g (approved):

martin-g:
Thank you, @x-paphio !

keesvandieren:
Thanks for fixing this.

Shouldn't Bootstrap Radios look like this: https://getbootstrap.com/docs/5.2/forms/checks-radios/

I think it will take some effort to get it aligned with the manual. Also, `ComponentsPage.html` is linking to old bootstrap docs, probably should link to https://getbootstrap.com/docs/5.3/forms/checks-radios/

x-paphio:
Hi. I know there is a difference there with the bootsrap doc. But I came to the understanding that the original template was based on bootswatch (https://bootswatch.com/cosmo/). So my contributation was to update following what bootswatch is doing for bootstrap 5.

PR #1152, opened 2024-10-08
Add missing caret to BootstrapSelectFilter by changing the css class

namtar:
When using the BootstrapSelectFilter the caret of the dropdown was not shown.
Changing the css class from "form-control" to "form-select" solves this issue (See: https://getbootstrap.com/docs/5.3/forms/select/).


martin-g (approved):

martin-g:
Thank you, @namtar !

PR #1175, opened 2024-12-20
Dependencies are updated; Javadoc warning were fixed; Failing test was disabled

solomax:


pedrosans:
Hi Maxim, 

In the work to fix the CVE-2024-53299, BasicResourceReferenceMapper was changed to only create ResourceReference for requests made for scope/name pairs matching a package resource. Because wicket-bootstrap-sass creates ResourceReference objects to resources in the servlet context instead, this test for sanity in the URL parameters fails and a ContextRelativeSassResourceReference is never created for the scope/name. I can:

1 - change the sanitization check to be extendable in wicket-core
2 - extend the sanitization check in wicket-bootstrap-sass to test for the ContextRelativeSassResourceReference#CONTEXT_RELATIVE_SASS_REFERENCE_VARIATION flag in the URL

Unfortunately I can't imagine a way to fix the failing test/issue in the current version.

pedrosans:
for reference: https://github.com/apache/wicket/pull/1060

martin-g:
Thank you, @solomax !

solomax:
@pedrosans @martin-g thanks for handling this one :)
Happy NY :))

PR #1214, opened 2025-05-02
Fix test dependency scopes

mweirauch:
While this change might break some users who haven't explicitely declared their test dependencies, these dependencies obviously shouldn't be in the compile/runtime classpath.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -288,6 +288,24 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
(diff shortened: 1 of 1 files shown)


martin-g:
Those should be set in the parent pom dependencyManagement

mweirauch:
> Those should be set in the parent pom dependencyManagement

Updated. Allthough I would personally always be explicit in the child module. Thanks for taking a look!

martin-g (approved):

martin-g:
Thank you, @mweirauch !

Issue #1252, opened 2025-10-01
wicket-bootstrap confirmation doesn't trigger `confirmed.bs.confirmation` JS event on parent element


solomax:
wicket-bootstrap confirmation doesn't trigger `confirmed.bs.confirmation` JS event on parent element

According to Bootstrap documentation (see for ex. here https://getbootstrap.com/docs/5.3/components/alerts/#events)

Bootstrap components MUST trigger events, this is now broken

@martin-g could you please wait with release a bit I'll try to fix this one :)))

martin-g:
Fixed with https://github.com/martin-g/wicket-bootstrap/pull/1253

PR #1253, opened 2025-10-01
Fixes issue #1252: bootstrap-confirmation client-side events are fixe…

solomax:
…d; bootstrap-confirmation.js code clean-up; Examples are updated

### DIFF
--- a/bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/pages/ExtensionsPage.java
+++ b/bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/pages/ExtensionsPage.java
@@ -3,11 +3,13 @@ package de.agilecoders.wicket.samples.pages;
 import java.time.Duration;
 import java.util.List;
 
+import org.apache.wicket.ajax.AjaxEventBehavior;
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.ajax.markup.html.AjaxLink;
 import org.apache.wicket.ajax.markup.html.form.AjaxButton;
 import org.apache.wicket.markup.head.CssHeaderItem;
 import org.apache.wicket.markup.head.IHeaderResponse;
+import org.apache.wicket.markup.head.OnDomReadyHeaderItem;
 import org.apache.wicket.markup.html.WebMarkupContainer;
 import org.apache.wicket.markup.html.basic.Label;
 import org.apache.wicket.markup.html.form.Form;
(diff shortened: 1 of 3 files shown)


bootstrap-samples/src/main/java/de/agilecoders/wicket/samples/pages/ExtensionsPage.java
```diff
@@ -334,9 +338,26 @@ protected void onSubmit(AjaxRequestTarget target) {
                 target.add(feedback);
             }
         };
+        /*confirmationButton.add(new AjaxEventBehavior("confirmed.bs.confirmation") {
```
solomax, line 341: Hello @martin-g, 

I probably found an issue in Wicket

for whatever reason this behavior is NOT working if being set before `ConfirmationBehavior`)

and works as expected if being set after ....

I was unable to find the reason and would appreciate any help with it :)))

martin-g: Hi @solomax !
Sorry! I don't have the time to help you with this.

reckart: I don't think it is a Wicket problem. If I see this correctly, AjaxEventBehavior and the ConfirmationBehavior both register a listener for "confirmed.bs.confirmation", so it is probably an issue boiling down to the order in which the event listeners are executed in the browser. I didn't check what exactly the listeners do though.

reckart: Correction: bootstrap-confimration.js doesn't listen for the event but trigger it. Still, are you sure it's not a client-side JS issue?

solomax: @reckart I wasn't able to find the root cause :((( tried to debug but with no luck :((

I'm also not sure it is wicket problem, but I can see no JS conflict, would appreciate if you can take a look :))

solomax: Since this potential issue is not affecting anyone else, maybe we can merge and release while I'll try to further investigate it in my spare time, WDYT? :))

martin-g:
@solomax Should I merge as-is for this release ?

solomax:
@martin-g yes, please :)

PR #1288, opened 2026-02-19
ClientSideBootstrapTabbedPanel css adjustments and tab change notification

namtar:
Hello,

I adjusted the css classes on the ClientSideBootstrapTabbedPanel so that they match Bootstrap 5.3.

Migration Guide for 5.3: https://getbootstrap.com/docs/5.3/migration/#navbars

* The .active class can no longer be applied to .nav-items, it must be applied directly on .nav-links.
* The .in class on the contents panel is now replaced by the .show class

The component already had the functionality to set an active tab at component creation (activeTabIndexModel). The tab change only happens on client side which leads to resetting the active tab to its initial value when refreshing the page via F5.

Additionally now one can activate the tab change notification via AJAX so that the activeTabIndexModel is updated. Refreshing the page via F5 keeps the current selected tab active.

Greetings


martin-g (approved):

martin-g:
Thank you, @namtar !

PR #1298, opened 2026-06-18
Close Modal Dialog with Escape Key

namtar:
Hi martin-g,

I discovered that the Modal does not close when pressing the "Escape "key because the browser ignores the event for the modal div. This is because the attribute "tabindex=-1" is missing.

As I have learned now, this attribute enables the modal div to have the keyboard focus set on it, because not all html components have this ability by default like <input>, ....

Independently of this fix I recognized, that the "keyboard" variable of de.agilecoders.wicket.core.markup.html.bootstrap.dialog.Modal has no effect because it is used nowhere and the "closeOnEscapeKey"-Variable does the job instead and initializes the bootstrap modal.js "keyboard" variable.

Attributes.set(tag, "data-bs-keyboard", "" + closeOnEscapeKey);

What to do? Should I delete the "keyboard" variable and its corresponding methods?

* de.agilecoders.wicket.core.markup.html.bootstrap.dialog.Modal#useKeyboard
* de.agilecoders.wicket.core.markup.html.bootstrap.dialog.Modal#setUseKeyboard

If I do this it could break the compilation of projects that use this methods. Can you tell me what is the best way to deal with this variable?

Greetings Namtar

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Modal.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/dialog/Modal.java
@@ -292,6 +292,10 @@ public class Modal<T> extends GenericPanel<T> {
             Attributes.removeClass(tag, "fade");
         }
 
+        // To receive the escape key event the tabindex attribute is needed. If it is missing the browser will ignore
+        // the event because the div has no focus. tabindex=-1 makes the modal div programmatically focusable
+        // but excludes it from the sequential keyboard navigation.
+        Attributes.set(tag, "tabindex", "-1");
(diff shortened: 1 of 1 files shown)


martin-g:
> If I do this it could break the compilation of projects that use this methods. Can you tell me what is the best way to deal with this variable?

The best would be to remove the field, deprecate the setter and the getter and they should just delegate to the setter/getter for `closeOnEscapeKey`.

martin-g:
Please also rebase your branch to latest `wicket-10.x-bootstrap-5.x` and force-push. Currently the PR says that you have 7 commits made 2 years ago.

namtar:
Thank you. I removed the field and deprecated the setter and getter and flagged them with forRemoval=true.

martin-g (approved):

martin-g:
Thank you, @namtar !

PR #1326, opened 2026-09-04
do not hide BootstrapGenericPanel's panels

michaelmosmann:
I had a problem where a a body panel had a null value inside its model, and this way was not rendered at all. Took me some time to find out why. I would expect that any provided component is rendered regardless of the containing model value.

Or did i miss something?

Thanks:)

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
@@ -114,17 +114,17 @@ public class BootstrapGenericPanel<T> extends GenericPanel<T>{
 		//Top Panel Image
         Component topImage = newTopImage(_PANEL_IMAGE_ID, getModel());
         add(topImage);
-        Components.hideIfModelIsEmpty(topImage);
+        Components.hideIfEmptyPanel(topImage);
 
 		//Panel Body
 		Panel panelBody = newBodyPanel(_PANEL_BODY_ID, getModel());
 		add(panelBody);
-		Components.hideIfModelIsEmpty(panelBody);
+		Components.hideIfEmptyPanel(panelBody);
 
 		//Panel Footer
 		Panel panelFooter = newFooterPanel(_PANEL_FOOTER_ID, getModel());
 		add(panelFooter);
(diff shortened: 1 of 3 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
```diff
@@ -114,17 +114,17 @@ protected void onInitialize()
 		//Top Panel Image
         Component topImage = newTopImage(_PANEL_IMAGE_ID, getModel());
         add(topImage);
-        Components.hideIfModelIsEmpty(topImage);
+        Components.hideIfEmptyPanel(topImage);
```
martin-g, line 117: I think we should remove the hiding altogether. It should be a decision of the specific implementation to decide what to do with its model.

martin-g (commented):
Hi @michaelmosmann ! :-)

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
michaelmosmann: This would be even better:)

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/panel/BootstrapGenericPanel.java
```diff
@@ -194,7 +191,7 @@ protected Panel newBodyPanel(String id, IModel<T> model){
 
 		Panel emptyPanel = new EmptyPanel(id);
 		emptyPanel.setDefaultModel(new Model<>());
-
+        emptyPanel.setVisible(false);
```
martin-g: Why is this needed ?
EmptyPanel renders nothing anyway - https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/markup/html/panel/EmptyPanel.html

michaelmosmann: you are right:) if nothing else depends on the visibility state, we can just remove this..  will do:)

martin-g (approved):

PR #1330, opened 2026-09-11
fix: animated progress bar 

michaelmosmann:
I think there were some kind of typo.. mixed active and animated .. but the bootstrap property is animated.. so the provided css class was wrong and the progress bar was not animated at all:)

https://getbootstrap.com/docs/5.3/components/progress/#animated-stripes

### DIFF
--- a/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/components/progress/ProgressBar.java
+++ b/bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/components/progress/ProgressBar.java
@@ -122,7 +122,7 @@ public class ProgressBar extends GenericPanel<Integer> {
 
         if (model != null) {
             Stack defaultStack = new Stack(getStackId(), () -> ProgressBar.this.getModelObject());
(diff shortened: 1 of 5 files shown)


bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/components/progress/Stack.java
```diff
@@ -91,11 +91,11 @@ public Stack striped(boolean value) {
         return this;
     }
 
-    public boolean active() {
```
martin-g, line 94: Let's keep the old names as deprecated forwarders, for easier migration.

bootstrap-core/src/main/java/de/agilecoders/wicket/core/markup/html/bootstrap/components/progress/ProgressBar.java
```diff
@@ -170,21 +170,21 @@ public ProgressBar striped(boolean value) {
         return this;
     }
 
-    public boolean active() {
+    public boolean animated() {
```
martin-g, line 181: deprecation here too

martin-g (approved):

